PR #2, opened 2019-02-06
Ported to Julia 0.7+

norrisryan:
I've updated everything to be compliant with Julia 0.7+.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-DustExtinction.jl
+# DustExtinction.jl
 =================
 
 [![Build Status](https://img.shields.io/travis/JuliaAstro/DustExtinction.jl.svg?style=flat-square)](https://travis-ci.org/JuliaAstro/DustExtinction.jl)
@@ -6,12 +6,20 @@ DustExtinction.jl
 
 Tools for interstellar dust extinction in astronomy
 
-Install
--------
+## Requirements
+------------
+As of Feb 5, 2018, Julia 0.7+ is required.
 
-```julia
-julia> Pkg.add("DustExtinction")
-```
+From a fresh Julia 0.7+ installation use the package manager (``` ]``` key) then do:
+
+``` add FITSIO```
+``` add Compat ```
+
+# Install
+-------
+From package manager (``` ]``` key) )
+
+``` add DustExtinction```
 
 Usage
 -----
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
 # DustExtinction.jl
-=================
 
 [![Build Status](https://img.shields.io/travis/JuliaAstro/DustExtinction.jl.svg?style=flat-square)](https://travis-ci.org/JuliaAstro/DustExtinction.jl)
 
@@ -7,17 +6,18 @@
 Tools for interstellar dust extinction in astronomy
 
 ## Requirements
-------------
+
 As of Feb 5, 2018, Julia 0.7+ is required.
 
 From a fresh Julia 0.7+ installation use the package manager (``` ]``` key) then do:
 
 ``` add FITSIO```
+
 ``` add Compat ```
 
-# Install
--------
-From package manager (``` ]``` key) )
+## Install
+
+From package manager (``` ]``` key)
 
 ``` add DustExtinction```
 
--- a/README.md
+++ b/README.md
@@ -5,16 +5,6 @@
 
 Tools for interstellar dust extinction in astronomy
 
-## Requirements
-
-As of Feb 5, 2018, Julia 0.7+ is required.
-
-From a fresh Julia 0.7+ installation use the package manager (``` ]``` key) then do:
-
-``` add FITSIO```
-
-``` add Compat ```
-
 ## Install
 
 From package manager (``` ]``` key)
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,2 @@
-julia 0.7
-FITSIO 0.13.0+
-Compat 1.4.0
+julia 1.0
+FITSIO 0.13.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
-julia 0.4
-FITSIO 0.8.1
-Compat 0.8.6
+julia 0.7
+FITSIO 0.13.0+
+Compat 1.4.0
--- a/src/DustExtinction.jl
+++ b/src/DustExtinction.jl
@@ -97,11 +97,11 @@ od94(w::Real, r_v) = ccm89like(w, r_v, od94_ca, od94_cb)
 # Vectorized versions (vectorized on wavelength only)
 for f = (:ccm89, :od94)
     @eval begin
-        ($f){T<:Real}(w::AbstractArray{T,1}, r_v) =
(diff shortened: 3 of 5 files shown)


REQUIRE
```diff
@@ -1,3 +1,3 @@
-julia 0.4
-FITSIO 0.8.1
-Compat 0.8.6
+julia 0.7
+FITSIO 0.13.0+
```
giordano: Why the `+`?  Isn't it for development version of packages?

test/runtests.jl
```diff
@@ -23,7 +23,8 @@ wave = 1.e4 ./ x_inv_microns
 ref_values = [1.569, 1.337, 1.000, 0.751, 0.479, 0.282, 0.190, 0.114]
 
 for i=1:length(wave)
-    @test_approx_eq_eps ccm89(wave[i], 3.1) ref_values[i] 0.016*ref_values[i]
+    #@test_approx_eq_eps ccm89(wave[i], 3.1) ref_values[i] 0.016*ref_values[i]
```
giordano: I think you can remove this line

test/runtests.jl
```diff
@@ -74,7 +75,8 @@ ref_values = [1.664, 1.321, 1.015, 0.819, 0.594,
               1.197, 0.811, 0.580]
 
 for i=1:length(wave)
-    @test_approx_eq_eps od94(wave[i], 3.1) ref_values[i] 0.0051*ref_values[i]
+    #@test_approx_eq_eps od94(wave[i], 3.1) ref_values[i] 0.0051*ref_values[i]
```
giordano: You can remove this line too

test/runtests.jl
```diff
@@ -115,7 +117,8 @@ if haskey(ENV, "SFD98_DIR")
     dustmap = SFD98Map()
     for i=1:length(refcoords)
         l, b = refcoords[i]
-        @test_approx_eq_eps ebv_galactic(dustmap, l, b) refebv[i] 0.02*refebv[i]
+        #@test_approx_eq_eps ebv_galactic(dustmap, l, b) refebv[i] 0.02*refebv[i]
```
giordano: And this line

src/DustExtinction.jl
```diff
@@ -252,17 +253,17 @@ function ebv_galactic(dustmap::SFD98Map, l::Real, b::Real)
         val = (1. - xw) * data[1] + xw * data[2]
     else
         data = read(hdu, x0:x0+1, y0:y0+1)
-        val = ((1.-xw) * (1.-yw) * data[1, 1] +
-               xw      * (1.-yw) * data[2, 1] + 
-               (1.-xw) * yw      * data[1, 2] +
+        val = ((1. -xw) * (1. -yw) * data[1, 1] +
```
giordano: I'd not use `1.`, better an explicit `1.0` (or probably `1` is even _more_ better)

README.md
```diff
@@ -1,17 +1,25 @@
...
 Tools for interstellar dust extinction in astronomy
 
-Install
--------
+## Requirements
 
-```julia
-julia> Pkg.add("DustExtinction")
-```
+As of Feb 5, 2018, Julia 0.7+ is required.
```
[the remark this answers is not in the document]
kbarbary: No need for a date here: older versions of Julia will still be able to install the package; they'll just get an older version.

giordano:
You have also to update the Travis script

README.md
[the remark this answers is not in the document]
norrisryan: Thank you for that point. I'll just remove the line about Julia version requirements.

README.md
```diff
@@ -1,17 +1,25 @@
...
+
+From a fresh Julia 0.7+ installation use the package manager (``` ]``` key) then do:
+
+``` add FITSIO```
```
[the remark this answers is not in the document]
norrisryan: If a user doesn't already have these packages, won't they have to add them? I thought it would be helpful to include instructions on how to do that.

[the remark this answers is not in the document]
giordano: No, the package manager takes care of it.  You don't have to manually add them, unless you want to explicit `using` them.

REQUIRE
```diff
@@ -1,3 +1,3 @@
...
+Compat 1.4.0
```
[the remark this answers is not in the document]
norrisryan: Thank you. This makes sense. I'll remove it from the requirements.

REQUIRE
norrisryan: To be honest, I just looked at the package number I was using. We were working with a development version of that package for different reasons. I'll fix that.

README.md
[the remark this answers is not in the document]
norrisryan: This could be the result of a limit in my experience with Julia. In DustExtinction.jl there is the line "using FITSIO". Won't users need to have added OIFITS then?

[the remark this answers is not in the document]
giordano: `FITSIO` is in the `REQUIRE` file, that's how the package manager knows that has to install `FITSIO` too :wink:

giordano:
@norrisryan a suggestion for you when working with Git: if you fork a repository and want to submit a PR against a branch (`master` in this case), create locally a new branch based on that one, don't use the same branch you eventually want to merge with :wink:

norrisryan:
Thank you for the patience and help. This my first PR for a package that was written outside of my research group. I've learned quite a bit!

giordano:
You're welcome, it's great to see new people interested in this project! :smiley:

src/DustExtinction.jl
```diff
@@ -150,8 +149,9 @@ directory containing the two FITS files defining the map,
 the FITS files defining the map open, speeding up repeated queries
 for E(B-V) values.
 """
-type SFD98Map
-    mapdir::Compat.UTF8String
+mutable struct SFD98Map
+    #mapdir::Compat.UTF8String depreciated
```
giordano: You can remove this line.  We don't need to keep old code, everything can be easily resurrected from git history, if needed :slightly_smiling_face:

src/DustExtinction.jl
```diff
@@ -150,8 +149,9 @@ directory containing the two FITS files defining the map,
...
+    mapdir::Core.String
```
giordano: Just use `String`, no need to prepend `Core.`

test/runtests.jl
```diff
@@ -23,7 +23,7 @@ wave = 1.e4 ./ x_inv_microns
 ref_values = [1.569, 1.337, 1.000, 0.751, 0.479, 0.282, 0.190, 0.114]
 
 for i=1:length(wave)
-    @test_approx_eq_eps ccm89(wave[i], 3.1) ref_values[i] 0.016*ref_values[i]
+    @test ccm89(wave[i], 3.1) ≈ ref_values[i] atol=0.016*ref_values[i]
```
giordano: Now that you're using the new `isapprox` syntax, you can simplify the condition by using the relative tolerance in place of the absolute tolerance: `rtol=0.016`.  You should do the same for the other tests.

norrisryan: Great! Thank you for the suggestions.

giordano:
Version 0.3.0 should now be available.  Thank you so much for your help, @norrisryan!  I'm sorry for having forgotten to merge for all these days.

PR #3, opened 2019-05-06
switched from old REQUIRE style to Project toml style

mileslucas:
To make sure project operates correctly, I used the REPL to run the tests:

```julia
(v 1.1) pkg> activate .
(DustExtinction) pkg> instantiate
(DustExtinction) pkg> test DustExtinction
Testing DustExtinction
 Resolving package versions...
Skipping SFD98Map test because $SFD98_DIR not defined.
Tests passed.
   Testing DustExtinction tests passed
```

I have also updated the version number (within `Project.toml`) to `0.3.1`. Feel free to change that.


Project.toml
```diff
@@ -0,0 +1,13 @@
+name = "DustExtinction"
+uuid = "fb44c06c-c62f-5397-83f5-69249e0a3c8e"
+license = "MIT"
+version = "0.3.1"
```
giordano: This should be `"0.3.1-DEV"`

Project.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[deps]
+FITSIO = "525bcba6-941b-5504-bd06-fd0dc1a4d2eb"
+
```
giordano: Please add a `compat` section:
```suggestion

[compat]
FITSIO = "≥ 0.13.0"
julia = "≥ 1.0.0"

```

giordano:
If you remove the `Manifest.toml` and ignore it, we're ready to go :wink:

mileslucas:
What is the purpose of ignoring the Manifest? Similar to `pipfile.lock` in python land, it does not have immediate use for general packaging, but for development it ensures consistency across packages.

giordano:
The `Manifests.toml` is a checkout of the current local state.  For me, it's great to have it locally and when you want to recreate an exact environment to run a code at a later point, but I see little value for having it checked in the repository of a package for development, shared between different people.

giordano:
Thanks!

PR #4, opened 2019-05-06
Update Travis script, and add AppVeyor and Cirrus scripts

giordano:


giordano:
I think I'll roll a new release of `FITSIO.jl` before merging this, in order to have support for FreeBSD

PR #5, opened 2019-05-06
Some bigger upgrades

mileslucas:
I've made a few changes and upgrades:

## 1. Refactoring
In general, I've refactored the extinction laws into their own files, as well as their corresponding tests. This is merely a style thing to reduce clutter in the main `DustExtinction.jl` and provide a better framework for adding additional laws. I've taken inspiration for this from the layout of `JuliaStats/HypothesisTests`

## 2. Broadcasting
Previously, all the functions were only applicable to scalars and arrays up to dim 2 with reshaping required with higher-level arrays. This involved that for-loop in `DustExtinction.jl` with `@eval` macros for basically aliasing the functions. Rather than using this, why not just use the `.` operator? E.g.:
```julia
julia> wave = 1e4:3e4
julia> mags = ccm89.(wave, r_v=3.1)
20001-element Array{Float64,1}:
 0.4039999999999999 
 0.4039349644872206 
 0.40386994594479847
[...]
```
This retains the same effects as the previous implementation, requires less code on our part, and stays idiomatic with the rest of the Julia mathematical syntax. This pull removes the previous definitions and changes all documentation to reference the `.` syntax.

## 3. Tests
I've added tests not only for the broadcasting but as well as the wavelength domain checking.

## 4. Documentation
I've added some documentation using `Documenter.jl`. It mostly mimics the README, but since this uses the docstrings it will be much more up to date than having to change the README everytime a function changes.

## 5. Calzetti 2000

I've added the Calzetti et al. (2000) extinction law!


giordano:
Could you please rebase your changes on `master`?  Also, it's not a good idea to work on a branch with the same name as the one you want to push to :wink:

giordano:
Another comment: please always end your files with a newline.  See [this](https://github.com/JuliaAstro/DustExtinction.jl/pull/5/files#diff-d443640aba44c5cb789f997af92abae4R7), [this](https://github.com/JuliaAstro/DustExtinction.jl/pull/5/files#diff-7ca30b2093d1972118e42e2abbf5db8eR13), [this](https://github.com/JuliaAstro/DustExtinction.jl/pull/5/files#diff-3b26734b5424473fbfb55c4a66866951R88), [this](https://github.com/JuliaAstro/DustExtinction.jl/pull/5/files#diff-03a40431280c4c7d9a938f0ef5b5a9a4R25), [this](https://github.com/JuliaAstro/DustExtinction.jl/pull/5/files#diff-a11d7e219ca62a3e854814507f0d5e6bR36), and [this](https://github.com/JuliaAstro/DustExtinction.jl/pull/5/files#diff-85ca2b58a19c3e93ffa31c97f99de5e1R105)

mileslucas:
Sorry about all the merge issues, I'm not quite sure how I managed to diverge so far!

I've also added newlines!

giordano:
I didn't look at the code yet, but by reading the description of the changes, they look great!  You need to rebase the changes so that we're able to accept the PR.  Probably merging from `origin/master` would help.

About vectorized functions, consider that this package was born way before the dot operator came into Julia, so manual action was required to make vectorization work (and this package was never updated to take advantage of the new feature).

mileslucas:
Okay, I think I've fixed the divergence, I realized I hadn't pulled from your master (I was only pulling from my fork's).

Project.toml
```diff
@@ -5,13 +5,14 @@ version = "0.3.1-DEV"
 
 [deps]
 FITSIO = "525bcba6-941b-5504-bd06-fd0dc1a4d2eb"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```
giordano: The package itself must not depend on `Documenter`.  See https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#.travis.yml-Configuration-1 about how to configure Travis to produce documentation

docs/src/index.md
```diff
@@ -0,0 +1,77 @@
...
+```
+(v 1.1) pkg> add DustExtinction
+[...]
+```
+
+## Usage
+
+###Color laws
+
+```julia
```
giordano: Use the `jldoctest` type, so that we can turn examples into actual tests and make sure they'll be updated.  See https://juliadocs.github.io/Documenter.jl/stable/man/doctests/  No need to do this in the previous block.

mileslucas:
Updated documentation according to the Documenter travis guide.

docs/Project.toml
```diff
@@ -0,0 +1,7 @@
+[deps]
+DustExtinction = "fb44c06c-c62f-5397-83f5-69249e0a3c8e"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+
+[compat]
+DustExtinction = ">=0.3.1-DEV"
+Documenter = "~0.20"
```
giordano: :slightly_smiling_face:

giordano: Probably this isn't clear if you don't visit the "Files changed" tab, but a line ending is missing here

giordano:
You also have to add the proper settings to the Travis script, otherwise documentation won't be built at all: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#.travis.yml-Configuration-1

.travis.yml
```diff
@@ -13,3 +13,13 @@ git:
   depth: 999999
 notifications:
   email: false
+
+jobs:
+include:
```
giordano: Indentation is important in YAML files, this should be
```yaml
  include:
    - stage: "Documentation"
      julia: 1.0
      os: linux
      script:
        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
                                               Pkg.instantiate()'
        - julia --project=docs/ docs/make.jl
      after_success: skip
```

giordano:
Yay, now documentation is being built: https://travis-ci.org/JuliaAstro/DustExtinction.jl/builds/528877796 :wink:

README.md
```diff
@@ -1,10 +1,14 @@
 # DustExtinction.jl
 
 [![Build Status](https://img.shields.io/travis/JuliaAstro/DustExtinction.jl.svg?style=flat-square)](https://travis-ci.org/JuliaAstro/DustExtinction.jl)
+[![](https://img.shields.io/badge/docs-stable-blue.svg)](https://JuliaAstro.github.io/DustExtinction.jl/stable)
+[![](https://img.shields.io/badge/docs-dev-blue.svg)](https://JuliaAstro.github.io/DustExtinction.jl/dev)
```
giordano: The link should be to https://JuliaAstro.github.io/DustExtinction.jl/latest, not dev

docs/Project.toml
```diff
@@ -0,0 +1,7 @@
+[deps]
+DustExtinction = "fb44c06c-c62f-5397-83f5-69249e0a3c8e"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+
+[compat]
+DustExtinction = ">=0.3.1-DEV"
```
giordano: Remove this line:
```suggestion
```

README.md
mileslucas: I copy and pasted those directly from the Documenter documentation, https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#Hosting-Documentation-1

giordano: Strange, I was looking to other docs build with Documenter in JuliaAstro:

* http://juliaastro.github.io/FITSIO.jl/latest/
* http://juliaastro.github.io/AstroLib.jl/latest/

Well, we can keep this and will see where the documentation goes

giordano: Uhm, probably newer versions of Documenter changed the link to `dev`

mileslucas:
Okay, I debugged the documentation, we won't get an actual link until we push to master

giordano:
Fine.  As I wrote above, probably they changed recently.  For `Measurements.jl` I have both latest and dev:
* http://juliaphysics.github.io/Measurements.jl/latest/
* http://juliaphysics.github.io/Measurements.jl/dev/

kbarbary:
I didn't review very carefully, but these all look like good changes. Two general comments:

- consider deprecation warnings for the `f(::Array, ...)` methods that must be changed to `f.(::Array, ...)`. Maybe there arent enough users that this is necessary.
- Would be interesting to re-run benchmarks and check that the dot syntax doesn't result in significant slowdowns (it shouldn't).

Nice work!

mileslucas:
Is there anything more desired from this pull request before merging?

giordano:
The deprecations asked by Kyle.  If you're not familiar with deprecations I can have a look at it

mileslucas:
I've added deprecations for `ccm89` and `od94` along with unit tests.

giordano:
Great job!

mileslucas:
Thanks, happy to help.

As a side note, the `stable` documentation branch will not work until a new tag or release is made. I've checked that `dev` does work, though. Do you think it is appropriate to make a new release now?

PR #6, opened 2019-09-01
Measurements and Unitful Support

mileslucas:
## Additions
1. Measurements.jl support

This actually worked right out of the box, but I added some units tests to the color laws to make sure uncertainty values work with them.

2. Unitful.jl support

I added a few extra methods of the form `law(wave::Quantity ...)` that convert incoming wavelengths to angstrom and output the calculated reddening in `mag`. If non-quantity units are input, non-quantity units are output. I've added unit tests for each color law to make sure they are compatible. 

For clarity:
```
law(wave::Float, ...) -> Float
law(wave::Quantity, ...) -> u"mag"
```

This adds 2 dependencies: Unitful and UnitfulAstro. 

## Testing
I added doctests directly to `test/runtests.jl` for an easier workflow. I ran into some issues with long numbers being displayed differently when I ran doctests through the test suite vs. through the `makedocs` command. What I've done is set `doctest=:fix` in `makedocs` which will fix the output of any doctests. I think this is OK because we are running the doctests in the testing suite already, so we still have coverage.  Any input on this would be great, otherwise we might just remove the one doctest causing this issue.

## Versioning
bumped from `0.4.0` to `0.5.0`. 

## Notes
The magnitude output is kind of buggy right now (see https://github.com/PainterQubits/Unitful.jl/issues/265) but it works for `erg/s` and `erg/s/cm^2` dimensions. Granted this project doesn't claim to actually do extinction, but rather just provide laws, it could be fine to let them output with `u"mag"` units and let the end-users handle them in their projects. If a fix gets pushed to Unitful or UnitfulAstro we'll have to pin the fixing version here as the lower bound.


mileslucas:
I think I overcomplicated things with the doctesting. I've gone back to just using doctests as part of the `makedocs` step and removed the doctest from the `runtests.jl` test suite. Hopefully all that will pass now.

Project.toml
```diff
@@ -1,17 +1,22 @@
 name = "DustExtinction"
 uuid = "fb44c06c-c62f-5397-83f5-69249e0a3c8e"
 license = "MIT"
-version = "0.4.0"
+version = "0.5.0"
 
 [deps]
 FITSIO = "525bcba6-941b-5504-bd06-fd0dc1a4d2eb"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
```
giordano: Maybe just require next version of `UnitfulAstro` (which will require `Unitful` 0.17 after https://github.com/JuliaAstro/UnitfulAstro.jl/pull/13 is merged)?  We don't directly need `Unitful` here, right?

docs/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
-DustExtinction = "fb44c06c-c62f-5397-83f5-69249e0a3c8e"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+DustExtinction = "fb44c06c-c62f-5397-83f5-69249e0a3c8e"
```
giordano: I should check documentation again, but I seem to remember that we shouldn't depend on the package itself here, because we want to use the local package that is in the local environment

Project.toml
mileslucas: Well, we do because we can’t access things like `ustrip` without ```using Unitful```. This means we need to explicitly add `Unitful` as a dep. Correct me if I’m wrong.

giordano: You're right, it's just me making confusion between the two pull requests :sweat_smile:

docs/Project.toml
mileslucas: You're right, it shouldn't be there!

PR #7, opened 2019-09-12
Dustmap update and extinct function

mileslucas:
This pull request achieves the following:

**1. Refactor dustmap code**
I've taken the existing dustmap code and refactored it to use the following syntax
```ebv_galactic(dustmap, l, b) -> dustmap(l, b)```
so usage now looks like
```julia
dmap = SFD98Map()
dmap(l, b)
# Vectorized
dmap.(ls, bs)
```
The previous `ebv_galactic` functions have been marked with a `@deprecated` macro.

**2. Added build step for SFD maps**
I've taken the download functions out of the code and put their bodies into `deps/build.jl`. Similar to `AstroLib.jl` this will download the maps during the build step into the `deps` folder by default. 

This takes the trouble out of the user's hands and the files are small enough to not cause significant issues with downloading (I think). If `SFD98_DIR` is set, the build step will download to there and the code will naturally read from there, so existing structure still exists.

**3. Added convenience `extinct` functions**
It may seem trivial, but since we've added first-class support for unitful, I think it's convenient to have functions that automatically apply reddening to fluxes. This is great because there is a natural divergence in the code between Unitful and normal values e.g.
```julia
extinct(f::Real, ...) = flux * 10 ^(-0.4 * Av * law(wave, Rv))
extinct(f::Quantity, ...) = flux * (Av * law(wave, Rv))
```

~Naturally, I've also added `extinct!` which only takes in flux values as an array and updates them in-place, although the existing `extinct` calls only work with scalars and use `.` to broadcast appropriately.~ See my comment below.

**4. Bump version to 0.6.0**


mileslucas:
Okay, I'm actually going to make some more changes. 

- [ ] Define `ColorLaw` type and create appropriate structs for each law
- [ ] Write out `(::ColorLaw)` methods for getting E(B-V) values
- [ ] Rewrite `redden` to take a law (think about approach for taking a map as well)
- [ ] Documentation and tests

giordano:
Changes to CI services should probably be factored out in a separate pull request

mileslucas:
Okay, I can do that.

mileslucas:
okay, I'm going to stop twiddling with stuff. Here's my update report:

**1. Type Stability**
I've added testing and made small code changes to ensure all of the code is type stable. Part of this included rewriting the `ccm89_invum` to use [`Polynomials.jl`](https://github.com/juliamath/polynomials.jl) `Poly` types. I have no clue of the performance impact since we've never done benchmarks before. For a future PR might be nice to test against Kyle Barbary's `extinction` python package which is implemented in cython. 

**2. Revised DustMap behavior**
As I wrote above, I've switched `SFD98Map` to have a call-function instead of using `ebv_galactic`, which has been tagged with a `@deprecate` macro. I've also ensured this dustmap is consistent with `Unitful.jl` and `Measurements.jl` support. 

**3. Refactor Files**
Moved files into a bit of a more logical tree with `color_laws.jl` and `dust_maps.jl`. I could imagine in the future implementing more functions (e.g. another color law) into `color_laws.jl` and if it grows too big, create a module for color laws and split up each one into its own file. If it makes sense, I can do that now, too.

**4. Much more documentation**
I've vamped up the documentation by a lot. I've gone through and revised most of the docstrings to ensure consistent wording, references, and examples. I've also added a little bit more background on what "Dust Extinction" is and why this tool is important. I've added plots, too :). See the above link for my most recent build.

**5. Version bump to `0.6.0`**
In regards to this, I think this project might be worthy of a major release if we can implement all the laws present in Kyle Barbary's `extinction` python package as well as what's available in astropy's `dust_extinction` package. Nobody ought to look at this package and say "it doesn't have what I need". 

** Bonus**
I've also added codecov since it was a simple flip of a switch.

CC @giordano this is ready for review.

.cirrus.yml
```diff
@@ -6,12 +6,11 @@ task:
     matrix:
       - JULIA_VERSION: 1.0
       - JULIA_VERSION: 1.1
+      - JULIA_VERSION: 1.2
```
giordano: Also Julia 1.3 is available

.travis.yml
```diff
@@ -1,26 +1,30 @@
 language: julia
+
 os:
   - linux
   - osx
+
 julia:
   - 1.0
   - 1.1
+  - 1.2
```
giordano: Also Julia 1.3 is available

README.md
```diff
@@ -35,21 +36,40 @@ julia> ccm89.([4000., 5000.], 3.1)
...
+ 1.5
+
+julia> extinct.(flux, wave, 0.3)
+2-element Array{Float64,1}:
+ 0.6671958182723856
+ 1.1000733242882896
+
+```
+
+We provide first-class support for `Unitful.jl` and `Measurements.jl` types, too! Check out the documentation for more examples.
```
giordano:
```suggestion
We provide first-class support for `Unitful.jl` and `Measurements.jl` packages, too! Check out the documentation for more examples.
```

appveyor.yml
```diff
@@ -1,7 +1,8 @@
 environment:
   matrix:
-  - julia_version: 1
+  - julia_version: 1.0
   - julia_version: 1.1
+  - julia_version: 1.2
```
giordano: Also Julia 1.3 is available

deps/build.jl
```diff
@@ -0,0 +1,18 @@
+# Downloads the SFD98 Map file
+const SFD98_BASEURL = "http://sncosmo.github.io/data/dust/"
+
+destdir = haskey(ENV, "SFD98_DIR") ? ENV["SFD98_DIR"] : @__DIR__
```
giordano: It shouldn't matter too much for performance here, but I think it's better to keep using the best practices when possible:
```suggestion
const destdir = haskey(ENV, "SFD98_DIR") ? ENV["SFD98_DIR"] : @__DIR__
```

src/color_laws.jl
```diff
@@ -0,0 +1,141 @@
...
+        k = 2.659 * (((0.011x - 0.198)x + 1.509)x - 2.156)
+    elseif x >  1 / 2.2
+        k = 2.659 * (1.040x - 1.857)
+    else
+        return 0.0x
+    end
+
+    return 1.0 + k / RV
+
+end
```
giordano: Missing newline at the end of this line

src/dust_maps.jl
```diff
@@ -0,0 +1,165 @@
...
+        crpix2 = dustmap.sgp_crpix2
+        lam_scal = dustmap.sgp_lam_scal
+        xsize, ysize = dustmap.sgp_size
+        n = -1
+    end
+
+    x, y = galactic_to_lambert(crpix1, crpix2, lam_scal, n, l, b)
+
+    # determine interger pixel locations and weights for bilinear interpolation
+    xfloor = floor(x)
```
giordano: Maybe you want to use `modf`?  If not, in case `x` and `y` can be negative probably `ceil` could be better than `floor`?

mileslucas: This was here from the original author. I'm not actually certain how it works, but I've tried out using `modf` and the tests still passed so I'll throw it in.

PR #8, opened 2019-09-30
Fixes to documentation and switch to using datadeps for the sfd map

mileslucas:
This will fix the current issues with docstrings and images on the documentation

- [x] Fix images not posting on hosted documentation
- [x] Fix dosctrings not rendering properly
- [x] Switch to using datadeps for the sfd map

This should be resolved and merged before we release `v0.6.0`.


mileslucas:
Problems have been fixed (huge thanks to @mortenpi for catching a unicode whitespace error). Ready for review.

PR #14, opened 2020-04-01
Fixed redundant broadcasting

### LINKED ISSUE #12: Redundant use of broadcasting
https://github.com/JuliaAstro/DustExtinction.jl/blob/a39125be77a98b8c397724199e3b389cb82e331f/src/color_laws.jl#L119

Here `λ` is already a scalar so I think broadcasting is redundant.

siddharthlal25:
Fixed #12.


mileslucas:
test failures seem unrelated to this, merging! Thanks for the catch.

PR #16, opened 2020-04-22
Added FM90 model

siddharthlal25:
This PR adds FM90 model for the package.


src/fittable_models.jl
```diff
@@ -0,0 +1,40 @@
...
+
+The parameters C1, C2, C3, C4, xo and gamma are y-intercept, slope, bump amplitude, FUV rise amplitude,
+bump centroid and bump width respectively.
+
+Fitzpatrick & Massa (1990) 6 parameter ultraviolet shape model, this model is only applicable at UV wavelengths.
+
+If `λ` is a `Unitful.Quantity` it will be automatically converted to Å and the
+returned value will be `UnitfulAstro.mag`.
+"""
+function FM90(λ::Real, C1 = 0.10, C2 = 0.70, C3 = 3.23, C4 = 0.41, xo = 4.60, γ = 0.9)
```
siddharthlal25, line 17: @giordano @mileslucas What would be your suggestion regarding these fittable models? Should I go for the current implemented version, or should I make `FM90` a struct and rewrite the function like `(z::FM90)(x::Real)`?

siddharthlal25:
@giordano I think we should remove Cirrus CI.

giordano:
Yes, Cirrus is not being helpful at all recently.

src/fittable_models.jl
mileslucas: This is actually a pretty big design choice. It makes sense to have a struct if a user fits it, then can evaluate many times without having to re-fit it. I would take that approach here, personally. 

To keep continuity, though, this implies we might want to change how the non-parametric models are presented. 

@giordano what do you think about having the following system?
- `struct LAW end`
- `(::LAW)(wave)`
- `function law() end`

where the last one is just a convenient alias?

giordano: Sounds good, let's see how it works

mileslucas: okay, I can follow the previous code up in a separate PR after this one to unify the interface. Will be a good opportunity to restructure some stuff, now that I have an extra ~6 months of Julia under my belt.

src/fittable_models.jl
```diff
@@ -0,0 +1,45 @@
...
+    C2::Real
+    C3::Real
+    C4::Real
+    xo::Real
+    γ::Real
+end
+
+FM90() = FM90(0.10, 0.70, 3.23, 0.41, 4.60, 0.9)
+
+function (z::FM90)(λ::Real)
```
siddharthlal25, line 27: Does this look good for the interface? @giordano @mileslucas

src/fittable_models.jl
```diff
@@ -0,0 +1,45 @@
...
+If `λ` is a `Unitful.Quantity` it will be automatically converted to Å and the
+returned value will be `UnitfulAstro.mag`.
+"""
+struct FM90
+    C1::Real
+    C2::Real
+    C3::Real
+    C4::Real
+    xo::Real
+    γ::Real
```
giordano, line 22: Please, pay attention to the types of the fields: they shouldn't be abstract types like `Real`.  Either use a fixed type (often too restrictive) or make the type parametric.

src/fittable_models.jl
giordano: Yeah, I think this is good.

siddharthlal25: I wanted to support `Quantity` as a form of input for lamda. I tried this:
```
function (z::FM90)(λ::Quantity) = (z)(ustrip(u"angstrom", λ)) * u"mag"
```
but it says `syntax: unexpected "="`. How to work this out?

giordano: Remove the `function` keyword :wink:

src/fittable_models.jl
```diff
@@ -0,0 +1,45 @@
...
+end
+
+FM90() = FM90(0.10, 0.70, 3.23, 0.41, 4.60, 0.9)
+
+function (z::FM90)(λ::Real)
+    x = aa_to_invum(λ)
+
+    if !(1.0 / 0.32 <= x <= 1.0 / 0.0912)
+        return zero(eltype(x))
```
mileslucas, line 31: This can cause type instability, because x may be an integer but everything we normally return is a float. For now `0.0x` has worked well.

src/fittable_models.jl
```diff
@@ -0,0 +1,45 @@
+# Convenience function for wavelength conversion
+aa_to_invum(wave::Real) = 10000 / wave
+aa_to_invum(wave::Quantity) = aa_to_invum(ustrip(u"angstrom", wave))
+
+"""
+    FM90(C1 = 0.10, C2 = 0.70, C3 = 3.23, C4 = 0.41, xo = 4.60, γ = 0.9)(λ::Real)
+
+The parameters C1, C2, C3, C4, xo and γ are y-intercept, slope, bump amplitude, FUV rise amplitude,
```
mileslucas, line 8: This is not a grammatically correct sentence.   Perhaps create a list instead
```
### Parameters
* `c1` - y-intercept
```
Etc. 

Are these the exact variable names used from the paper? They’re awfully not indicative of what they do. We also may want to set this up using Parameters.jl for kwdef the struct. What do you think @giordano ?

src/fittable_models.jl
giordano: As an alternative
```suggestion
        return zero(float(eltype(x)))
```
could give the zero of the corresponding float type

src/fittable_models.jl
giordano: Yes, I was thinking about using `Base.@kwdef`, but if this is going to be a parametric type I think that this macro supports them only in Julia v1.1+ (better double check this), so `Parameters.jl` would be a better option in that case.

siddharthlal25: I referred to astropy codebase. :sweat_smile: I am switching to the list format.
Sorry about the grammatical mistake.

src/fittable_models.jl
siddharthlal25: Okay, I will fix it. Do you have any good resource to learn about type stability and how to fix instabilities? @giordano @mileslucas

mileslucas: Mostly just practice. There’s a few tools for testing stuff- @code_warntype, @inferred, Cthulu.jl

I think the harder question is knowing when you absolutely need to apply and and when it’s not worth the trouble.

src/fittable_models.jl
mileslucas: Don’t sweat the grammar! If you can find the parameter names from the paper they might have better names, or perhaps worse. I just want to minimize the _need_ for documentation to understand what those values are.

mileslucas: Also, base kwdef doesn’t exist in 1.0, so we should use Parameters.jl. They also have a neat syntax for parameteized and default values struct, just like this!

giordano: > base kwdef doesn’t exist in 1.0

I think it does, but it doesn't support parametric structs, as I said above:

```
help?> Base.@kwdef
  @kwdef typedef

  This is a helper macro that automatically defines a keyword-based constructor for the type declared in the expression typedef, which must
  be a struct or mutable struct expression. The default argument is supplied by declaring fields of the form field::T = default or field =
  default. If no default is provided then the keyword argument becomes a required keyword argument in the resulting type constructor.

  Inner constructors can still be defined, but at least one should accept arguments in the same form as the default inner constructor (i.e.
  one positional argument per field) in order to function correctly with the keyword outer constructor.

  │ Julia 1.1
  │
  │  Base.@kwdef for parametric structs, and structs with supertypes requires at least Julia 1.1.
```

src/fittable_models.jl
siddharthlal25: Would the function offer better performance if written as `function (z::FM90)(x::T) where T <: Number`?

mileslucas: Nope. Typing is for dispatch only, it won’t make your code faster. Also, in general, if you want to ask “is it better performance” is really hard to judge from visual inspection only. Using BenchmarkTools.jl and playing around yourself will be far more useful.

giordano:
Please, rebase on/merge `master`

src/fittable_models.jl
giordano: To add to what Miles said, you may be interested in reading https://white.ucc.asn.au/2020/04/19/Julia-Antipatterns.html

siddharthlal25:
@giordano @mileslucas I would like to draw your attention towards the current design and structure of the code and would like to know your suggestions regarding the code? If it seems good to you guys, I will proceed with implementing more models following the same code structure. :smiley:

mileslucas:
It needs a design update, but that can be addressed in a different PR. Also we should keep these PRs minimal (don’t introduce too many new things at once, for example). 

For now I think let’s wrap to this PR and we can figure out the new design moving forward after that.

siddharthlal25:
Okay! So let's keep this PR to FM90 model only and let's start the design update before adding more model. Does that sound good? @giordano @mileslucas

src/fittable_models.jl
siddharthlal25: Thank you so much for all the help!

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.6.0"
 [deps]
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
 FITSIO = "525bcba6-941b-5504-bd06-fd0dc1a4d2eb"
+Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
```
mileslucas, line 9: make sure to add a Compat entry, too

src/fittable_models.jl
```diff
@@ -0,0 +1,59 @@
...
+    C4::T = 0.41
+    xo::T = 4.60
+    γ::T = 0.99
+
+    function FM90{T}(C1::T, C2::T, C3::T, C4::T, xo::T, γ::T) where T <: Number
+        xo < 0 && error("Invalid axis xo=$xo. xo must be greater than or equal to 0")
+        γ < 0 && error("Invalid axis γ=$γ. γ must be greater than or equal to 0")
+        new{T}(C1, C2, C3, C4, xo, γ)
+    end
+end
```
mileslucas, line 35: this can be tidied up (from [Parameters.jl manual](https://mauro3.github.io/Parameters.jl/stable/manual/))

I've also found the derivation in the original paper. The parameters they use are pretty simple- there are four terms, each with a coefficient. there is a constant term (c1), a linear term (c1 * x), a Drude profile (c3 * D(x; gamma, x0)) and a "curvature" term (c4 * F). Given this, as well as the consideration that we don't want to force our users into using special unicode characters (like \gamma as a keyword arg), I think we should parametrize the struct with
`c1,c2,c3,c4,x0,gamma`.


```julia
@with_kw struct FM90{T<:Number} @deftype T
    c1 = 0.10
    c2 = 0.70
    c3 = 3.23
    c4 = 0.41
    x0 = 4.60
    gamma = 0.99
    @assert x0 \geq 0 "`x0` must be \geq 0, got $x0"
    @assert gamma \geq 0 "`gamma` must be \geq 0, got $gamma"
end
```

src/fittable_models.jl
```diff
@@ -51,4 +51,4 @@ function (z::FM90)(λ::Real)
     return exvebv
 end
 
-(z::FM90)(λ::Quantity) = z(ustrip(u"angstrom", λ)) * u"mag"
+(z::FM90)(λ::Quantity) = z(ustrip(u"angstrom", λ))
```
mileslucas, line 54: Why? This makes the outputs no longer unitful.

siddharthlal25: Fixed it!

src/fittable_models.jl
```diff
@@ -0,0 +1,54 @@
...
+* `c2` - slope of liner term
+* `c3` - amplitude of 2175 Å bump
+* `c4` - amplitude of FUV rise
+* `x0` - centroid of 2175 Å bump
+* `gamma` - width of 2175 Å bump
+
+Fitzpatrick & Massa (1990) 6 parameter ultraviolet shape model, this model is only applicable at UV wavelengths.
+
+If `λ` is a `Unitful.Quantity` it will be automatically converted to Å and the
+returned value will be `UnitfulAstro.mag`.
```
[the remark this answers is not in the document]
mileslucas: note the above code blocks are wrapped in double graves "``", so they show up as LaTeX in Documenter, but show up as code blocks on GitHub

src/fittable_models.jl
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    exvebv = z.c1 + z.c2 * x
+
+    x2 = x^2
+    exvebv += z.c3 * (x2 / ((x2 - z.x0^2)^2 + x2 * (z.gamma^2)))
+
+    if x >= 5.9
+        y = x - 5.9
+        exvebv += z.c4 * @evalpoly y 0.0 0.0 0.5392 0.05644
```
mileslucas, line 48: hm, I'm not sure which is better
```
exebv += z.c4 * @evalpoly y 0.0 0.0 0.5392 0.05644
```
```
exebv += z.c4 * (0.5392y^2 + 0.05644y^3)
```

siddharthlal25: The first one outperforms the second for large values of y.

src/fittable_models.jl
```diff
@@ -0,0 +1,54 @@
+
+
+"""
+    FM90(c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Real)
+    FM90(c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Quantity)
```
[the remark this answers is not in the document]
siddharthlal25: Should I keep it like this:
```
FM90(;c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Real)
FM90(;c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Quantity)
FM90(coeffs, x0=4.60, gamma=0.9)(λ::Real)
FM90(coeffs, x0=4.60, gamma=0.9)(λ::Quantity)
```

or this:
```
FM90(;c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Real)
FM90(coeffs, x0=4.60, gamma=0.9)(λ::Real)
```

or this:
```
FM90(;c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Real)
FM90(coeffs, x0=4.60, gamma=0.9)(λ::Quantity)
```
?

src/fittable_models.jl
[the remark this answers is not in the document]
siddharthlal25: I am really sorry about my documenting skills :sweat_smile:, I will improve on it!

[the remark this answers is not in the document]
mileslucas: No problem! It's hard to learn how to write good documentation (and I don't claim to be great at it either), but it helps that I have a little more experience with the domain this is from.

src/fittable_models.jl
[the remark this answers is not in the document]
mileslucas: Perhaps my comment wasn't clear: you need to change the docstring only, and actually show the available methods. What I typed up are the two methods you should show in the docstring

src/fittable_models.jl
mileslucas: we don't expect y to be large, though. Consider that `x` is restricted from `3.3, 8.0` already, and `y = x - 5.9`, so the largest absolute value is `2.1`

siddharthlal25: I used `@btime` and wrapped the above two as functions, it resulted in approximately the same time and allocation values for both of them (for small values of `y`). So I think we should keep the first one for code uniformity. What would you suggest?

mileslucas: sounds like reasonable logic to me!

src/fittable_models.jl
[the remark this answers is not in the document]
siddharthlal25: Should we keep the parameters table or drop it?

docs/src/fitting_models.md
```diff
@@ -0,0 +1,105 @@
...
+
+julia> FM90().(wave_new)
+5-element Array{Gain{Unitful.LogInfo{:Magnitude,10,-2.5},:?,Measurement{Float64}},1}:
+     5.15 ± 0.12 mag
+   4.809 ± 0.062 mag
+   4.743 ± 0.029 mag
+ 4.9684 ± 0.0036 mag
+     5.66 ± 0.63 mag
+
+```
```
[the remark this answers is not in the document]
siddharthlal25: Added the plot!

docs/src/fitting_models.md
```diff
@@ -0,0 +1,105 @@
...
+
+julia> wave = [1000, 1200, 1800];
+
+julia> model.(wave)
+3-element Array{Float64,1}:
+ 12.562237969522851
+  7.769215017329513
+  4.890128210972148
+
+```
```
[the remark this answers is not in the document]
siddharthlal25: Shifted a few under `# Examples` section of `FM90` model.

docs/make.jl
```diff
@@ -11,7 +11,8 @@ makedocs(modules = [DustExtinction],
     pages = [
         "Home" => "index.md",
         "Color Laws" => "color_laws.md",
-        "Dust Maps" => "dust_maps.md"
+        "Dust Maps" => "dust_maps.md",
+        "Fitting Models" => "fitting_models.md"
```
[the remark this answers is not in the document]
siddharthlal25: I removed a separate page for "Fitting Models" and added it under `color_laws`.

docs/src/fitting_models.md
```diff
@@ -0,0 +1,105 @@
...
+
+julia> model(1500)
+5.2521258452800135
+
+julia> FM90()(1500)
+5.152125845280013
+
+```
+
+These models can be applied across higher dimension arrays using the `.` operator
```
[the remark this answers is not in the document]
siddharthlal25: Can you give me some technical definition which people in the astronomy community use about the "Fitting Models"?

docs/src/fitting_models.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+```jldoctest
+julia> model = FM90(c1 = 0.2, c2 = 0.7, c3 = 3.23, c4 = 0.41, x0 = 4.6, gamma = 0.99);
+
+julia> model(1500)
+5.2521258452800135
+
+julia> FM90()(1500)
+5.152125845280013
```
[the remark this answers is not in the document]
siddharthlal25: Placed it in the example section!

docs/src/fitting_models.md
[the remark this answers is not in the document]
siddharthlal25: Also removed this page of extra documentation since it was repetitive.

docs/make.jl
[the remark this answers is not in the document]
siddharthlal25: Currently, I am using the name "Fittable Extinction Laws" in the documentation part, let me know if it is good or should be changed.

src/fittable_models.jl
```diff
@@ -0,0 +1,79 @@
+
+
+"""
+    FM90(;c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Real)
+    FM90(;c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Quantity)
+    FM90(coeffs, x0=4.60, gamma=0.9)(λ::Real)
+    FM90(coeffs, x0=4.60, gamma=0.9)(λ::Quantity)
```
giordano, line 7: To avoid duplications we can have
```suggestion
    FM90(;c1 = 0.10, c2 = 0.70, c3 = 3.23, c4 = 0.41, x0 = 4.60, gamma = 0.9)(λ::Union{Real, Quantity})
    FM90(coeffs, x0=4.60, gamma=0.9)(λ::Union{Real, Quantity})
```

mileslucas: I agree, I also find the spaces around every = quite disruptive and overextend the length of the signature.

siddharthlal25: I will remove the spaces in next commit after we decide about the parameter tally and the name of the new category.

mileslucas:
Thanks so much for this! I'm going to merge and update with our new interface and release. In the future, take a look at the docs I've written for `ExtinctionLaw`, they explain what needs to happen for future laws!

PR #22, opened 2020-05-03
Removed the Polynomials dependency

siddharthlal25:
Initially, there was only one function `ccm_invum` which was used for both `ccm89` and `od94` models. The only point of difference in these models was polynomial evaluations, so the polynomials were explicitly passed to the `ccm_invum` which lead to Polynomials as a dependency.
This PR creates two new functions (`ccm_invum` & `od94_invum`) to address this issue and in polynomial evaluation, `@evalpoly` is used.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,14 +6,12 @@ version = "0.6.0"
 [deps]
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
 FITSIO = "525bcba6-941b-5504-bd06-fd0dc1a4d2eb"
-Polynomials = "f27b6e38-b328-58d1-80ce-0feddd5e7a45"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 UnitfulAstro = "6112ee07-acf9-5e0f-b108-d242c714bf9f"
 
 [compat]
 DataDeps = "0.7"
 FITSIO = "0.13.0, 0.14"
-Polynomials = "0.8"
 Unitful = "0.17.0, 1"
 UnitfulAstro = "0.3.0, 0.4, 1"
 julia = "1.0.0"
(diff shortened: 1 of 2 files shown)


siddharthlal25:
Looking at the codecov results for `od94` (https://codecov.io/gh/JuliaAstro/DustExtinction.jl/pull/22/changes#L104), it seems that it has not been tested properly. I will be adding some more tests once this PR is wrapped up.

mileslucas:
Why write two functions? Before polynomials we just passed in vectors of coefficients.

siddharthlal25:
We can definitely pass the coefficients as vectors, but since the code for subroutine `invum` of both the models are the same so we can't get the real code coverage stats.
Like right now when I wrote two different functions for `invum` for the models, I got the stat that `od94` model was not tested completely. So, I think we should keep them separate, although the volume of code increases but we get a clear picture of the testing of functions and also the readability as well as uniformity of code increases. What would you suggest?

mileslucas:
Following code coverage does not imply good or bad tests. It only implies the code has been run. 

The od94 algorithm is specifically the same as ccm89 except for those coefficients. That was how the paper was written and it’s clear this should lead to code reuse. 

So, consider I write only the ccm89 function, and I implement many tests. I notice through code coverage that one of the if-else branches hasn’t been tested. So I add some tests for accuracy. At this point, I have all the testing I need to put this function out on the wild. 

Now, I write the od94 function, building off my previous one. And I wrote some tests that make sure the _differences_ are well tested. 

Code coverage says in both cases there’s 100% coverage. When you split them apart, the code for od94 doesn’t look as “well-tested” because of less code coverage for it. What’s going on? Well, since od94 is just a marginal change to ccm89, why should I have to test the shared code twice?

Moral of the story is: od94 is a marginal change to ccm89, so if we have tests for the shared functionality, it doesn’t matter where it was called.

siddharthlal25:
The method adopted to remove Polynomials dependency:
Coefficients of polynomials were passed as vector and the coefficients were explicitly written for polynomial evaluation as `@evalpoly y c_a[1] c_a[2] ... c_a[9]`.

PR #25, opened 2020-05-11
Redo the interface

### LINKED ISSUE #23: Redundant utility function
The general format of a model is of the form: `model(λ::Real, Rv = 3.1)`
and if `λ` is `Quantity`, then we have a modified definition to the function call being:
`model(λ::Quantity, Rv::Real = 3.1) = model(ustrip(u"angstrom", λ), Rv) * u"mag"`.

Therefore function `aa_to_invum` in each of these models will be called with `λ::Real` and not as `λ::Quantity`. Therefore the utility function mentioned below is redundant.
https://github.com/JuliaAstro/DustExtinction.jl/blob/1766089ca9b10a7892684509df3aff0e7b85ceee/src/color_laws.jl#L5

mileslucas:
Big API change for better facilitating the extension and common interface for dust laws.

## New things
`ExtinctionLaw` abstract type

struct + "functor" (what are these things called, for real? I don't know what the official term is)
`ccm89(wave, Rv)` -> `CCM89(Rv=Rv)(wave)`
* through this we create an interface for the parametrized laws that is equivalent to potentially fittable laws, and would allow storing laws easier in other structs. 
* For the parametrization, I stuck with using Parameters.jl and kwdeffing with `Float64` for the Rv type. I could make it abstract and put into the struct parameters, but I'm not sure that's really worth it for such a thin-wrapper
* This will allow us to easily add a plot recipe for the laws, if we want, in combination with the `bounds` function like `plot(CCM89())` to get the standard magnitudes and like `plot(CCM89(), :invum)` to plot using inverse microns for the x-axis

## Changes
So far I'm trying to deprecate everything; I've got no clue what the user-base for this is, but we ought to deprecate everything until we get a little more stable and are ready to release a v1.0. I've started putting these into `deprecate.jl` with annotations for convenience

`redden` and `deredden` now have a different interface- the new docstrings show it best, but basically you call
```julia
redden(law, wave, flux; Av=1, law_kwargs...)
```
where `law` can be either a type `CCM89` (which will then pass in the `law_kwargs` to the constructor) or a fully constructed `ExtinctionLaw`
```julia
redden(CCM89, 3000, 1000) # Rv=3.1 by default
redden(CCM89, 3000, 1000; Rv=3.1)
redden(CCM89(Rv=3.1), 3000, 1000)
redden(CCM89(3.1), 3000, 1000)
redden(CCM89(), 3000, 1000)
```
all return the same thing, which is kind of amazing consider how little code it takes to make that interface.

I've inserted a bit of code-gen to minimize what needs to be written for each law, right now it's just handling the unitful conversions. 

----

Right now tests are broken and I haven't dove into the docs yet, and I'm sleepy, but this is my current working concept for a more usable interface, I'd love to hear comments.

Edit: Docs preview ought to land [here](https://juliaastro.github.io/DustExtinction.jl/previews/PR25/index.html)

Fixes #23

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DustExtinction"
 uuid = "fb44c06c-c62f-5397-83f5-69249e0a3c8e"
 license = "MIT"
-version = "0.6.0"
+version = "0.7.0"
 
 [deps]
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
--- a/Project.toml
+++ b/Project.toml
@@ -6,12 +6,14 @@ version = "0.6.0"
 [deps]
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
 FITSIO = "525bcba6-941b-5504-bd06-fd0dc1a4d2eb"
+Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 UnitfulAstro = "6112ee07-acf9-5e0f-b108-d242c714bf9f"
 
 [compat]
 DataDeps = "0.7"
 FITSIO = "0.13.0, 0.14"
+Parameters = "0.12"
 Unitful = "0.17.0, 1"
 UnitfulAstro = "0.3.0, 0.4, 1"
 julia = "1.0.0"
--- a/docs/src/color_laws.md
+++ b/docs/src/color_laws.md
@@ -1,7 +1,7 @@
 
 # [Color laws](@id laws)
 
-The following empirical laws allow us to model the reddening of light as it travels to us. The law you use should depend on the type of data you have and the goal of its use. [`ccm89`](@ref) is very common for use in removing extinction from stellar observations, but [`cal00`](@ref), for instance, is suited for galaxies with massive stars. Look through the citations and documentation for each law to get a better idea of what sort of physics it targets.
+The following empirical laws allow us to model the reddening of light as it travels to us. The law you use should depend on the type of data you have and the goal of its use. [`CCM89`](@ref) is very common for use in removing extinction from stellar observations, but [`CAL00`](@ref), for instance, is suited for galaxies with massive stars. Look through the citations and documentation for each law to get a better idea of what sort of physics it targets.
 
 ```@meta
 DocTestSetup = quote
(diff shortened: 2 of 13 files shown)


Project.toml
```diff
@@ -1,17 +1,19 @@
 name = "DustExtinction"
 uuid = "fb44c06c-c62f-5397-83f5-69249e0a3c8e"
 license = "MIT"
-version = "0.6.0"
+version = "0.7.0"
```
giordano, line 4: I may have already asked this in the past, but how about going to v1.0?

src/DustExtinction.jl
```diff
@@ -1,53 +1,132 @@
...
+# --------------------------------------------------------------------------------
+# Here be codegen!
+
+# generate unitful support for the following laws
+# this can be removed when julia support is pinned to 1.3 or higher,
+# at which point adding `(l::ExtinctionLaw)(wave)` is possible, until then
+# using this code-gen does the trick but requires manually editing
+# instead of providing support for all <: ExtinctionLaw
+for law in [:CCM89, :OD94, :CAL00, :GCC09, :VCG04]
+    @eval (l::$law)(wavelength::Quantity) = l(ustrip(u"Å", wavelength)) * u"mag"
```
giordano, line 126: Why not simply
```suggestion
for law in [CCM89, OD94, CAL00, GCC09, VCG04]
    (l::law)(wavelength::Quantity) = l(ustrip(u"Å", wavelength)) * u"mag"
```
?  The `@eval` macro is not really needed

docs/src/color_laws.md
```diff
@@ -43,20 +45,17 @@ julia> wave = range(4000, 5000, length=4)
...
-```
+julia> deredden.(CCM89(Rv=3.1), wave, ans; Av=0.3) ≈ flux
+true
 
-!!! note
-    If you are concerned with memory and would like an in-place broadcasted version of [`redden`](@ref) use it with [`Base.map!`](https://docs.julialang.org/en/v1/base/collections/#Base.map!)
-    ```julia
-    map!((f,w)->redden(f, w, 0.3), flux, flux, wave)
-    ```
+```
```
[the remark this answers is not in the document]
mileslucas: Haha it’s nonsense isn’t it? The same thing is just
```
@. flux = redden(LAW, wave, flux; kwargs...)
```

Project.toml
mileslucas: I think with this change we should be pretty well setup to facilitate any new laws, but let’s sit on the interface a little bit longer before committing. At 1.0 I assume we would drop all the depredations we’ve been holding. I wish I were using this more actively so I could find the pain points.

src/DustExtinction.jl
mileslucas: Because meta programming always starts out as “okay I got this” and always ends with “don’t use @eval” and I never learn my lesson 😂

docs/src/color_laws.md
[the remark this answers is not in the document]
giordano: I see, right!

Issue #10, opened 2020-02-27
Implement More Methods!


mileslucas:
Here are a list of dust extinction models that have been implemented somewhere else (along with their paper) that could be implemented into this repository.

- [x] ccm89 [code](https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.CCM89.html#dust_extinction.parameter_averages.CCM89) (https://github.com/JuliaAstro/DustExtinction.jl/commit/ae9ed2555c652d1c1690bfe6072f97a283bb4112)
- [x] od94 [code](https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.O94.html#dust_extinction.parameter_averages.O94) (https://github.com/JuliaAstro/DustExtinction.jl/commit/ae9ed2555c652d1c1690bfe6072f97a283bb4112)
- [x] cal00 [code](https://dust-attenuation.readthedocs.io/en/latest/api/dust_attenuation.averages.C00.html) (https://github.com/JuliaAstro/DustExtinction.jl/pull/5/commits/5e5011bbf0d9feb9b12246dcc78027cdc336442d)
- [x] f99 [code](https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.F99.html#dust_extinction.parameter_averages.F99) (#30)
- [x] f04 [code](https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.F04.html#dust_extinction.parameter_averages.F04) (#31)
- [x] vcg04 [code](https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.VCG04.html#dust_extinction.parameter_averages.VCG04) (#13 thanks @siddharthlal25 )
- [x] gcc09 [code](https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.GCC09.html#dust_extinction.parameter_averages.GCC09) (#13 thanks @siddharthlal25 )
- [x] m14 [code](https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.M14.html#dust_extinction.parameter_averages.M14) (#33)
- [x] g16 [code](https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.G16.html#dust_extinction.parameter_averages.G16) (#34)
[...]

PR #29, opened 2020-07-04
Added P92 model

siddharthlal25:
Adding P92 model, we can mention this in #10!


src/fittable_laws.jl
```diff
@@ -77,3 +77,135 @@ function (law::FM90)(wave::T) where T
...
+julia> P92(FUV_b = 2.0).([1000, 2000, 3000])
+3-element Array{Float64,1}:
+ 3.8390886792306187
+ 2.7304534614548697
+ 1.806181164464396
+
+```
+
+## References
+[Pei (1992)](https://ui.adsabs.harvard.edu/abs/1992ApJ...395..130P)
```
mileslucas, line 140: I would add an extra section that has a table with the default parameters and the bandpass, something like

| | lambda | A | b | n |
| - | - | - | - | - |
| FIR | | | | |

siddharthlal25:
@mileslucas @giordano it would be really nice if you can review this!

giordano (approved):

src/fittable_laws.jl
```diff
@@ -77,3 +77,152 @@ function (law::FM90)(wave::T) where T
...
 end
 
+
+"""
+    P92(BKG_amp=218.57142857142858, BKG_lambda=0.047, BKG_b=90.0, BKG_n=2.0, FUV_amp=18.545454545454547, FUV_lambda=0.07,
+        FUV_b=4.0, FUV_n=6.5, NUV_amp=0.05961038961038961, NUV_lambda=0.22, NUV_b= -1.95, NUV_n=2.0, SIL1_amp=0.0026493506493506496,
+        SIL1_lambda=9.7, SIL1_b=-1.95, SIL1_n=2.0, SIL2_amp=0.0026493506493506496, SIL2_lambda=18.0, SIL2_b= -1.8, SIL2_n=2.0,
+        FIR_amp=0.015896103896103898, FIR_lambda=25.0, FIR_b=0.0, FIR_n=2.0)
+
+Pei (1992) generative model applicable from the extreme UV to far-IR.
```
mileslucas, line 87:
```suggestion
Pei (1992) generative extinction model applicable from the extreme UV to far-IR.
```

mileslucas:
Also, you are missing an entry in the table in `docs/src/index.md` corresponding to the bibliography entry

src/fittable_laws.jl
```diff
@@ -77,3 +77,152 @@ function (law::FM90)(wave::T) where T
     return exvebv
 end
 
+
+"""
+    P92(BKG_amp=218.57142857142858, BKG_lambda=0.047, BKG_b=90.0, BKG_n=2.0, FUV_amp=18.545454545454547, FUV_lambda=0.07,
+        FUV_b=4.0, FUV_n=6.5, NUV_amp=0.05961038961038961, NUV_lambda=0.22, NUV_b= -1.95, NUV_n=2.0, SIL1_amp=0.0026493506493506496,
+        SIL1_lambda=9.7, SIL1_b=-1.95, SIL1_n=2.0, SIL2_amp=0.0026493506493506496, SIL2_lambda=18.0, SIL2_b= -1.8, SIL2_n=2.0,
+        FIR_amp=0.015896103896103898, FIR_lambda=25.0, FIR_b=0.0, FIR_n=2.0)
```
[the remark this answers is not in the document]
siddharthlal25: How about this?
Screenshot from 2020-07-20 06-10-12

src/fittable_laws.jl
```diff
@@ -77,3 +77,172 @@ function (law::FM90)(wave::T) where T
...
+    SIL1_n = 2.0
+    SIL2_amp = 0.002 * (1 / 3.08 + 1)
+    SIL2_lambda = 18.0
+    SIL2_b = -1.80
+    SIL2_n = 2.0
+    FIR_amp = 0.012 * (1 / 3.08 + 1)
+    FIR_lambda = 25.0
+    FIR_b = 0.00
+    FIR_n = 2.0
+    @assert BKG_amp ≥ 0 "`BKG_amp` must be ≥ 0, got $BKG_amp"
```
mileslucas, line 204: I'm almost wondering if it would be easier to represent this struct hierarchically with tuples or named tuples
```julia
struct Example
	BKG = (165.0 * (1/3.08 + 1), 0.047, etc...)
    # or
    FIR = (A=0.012 * (1 / 3.08 + 1), lambda=25.0, b=0.0
end
```

It's clear this struct has this hierarchical order and it seems organized a little better. The question is whether this is easier for the user or if it makes sense for the user. Perhaps we don't use this for the struct itself, but we add a constructor that allows this behavior, something like
```julia
P92(BKG, FUV, NUV, SIL1, SIL2, FIR) = P92(BKG..., FUB..., NUV..., SIL1..., SILL2..., FIR...)
```
We could make this a kwarg-only version, but we'd have to remove the `@with_kw` from the struct (which is fine, we'd just have to add the assertion checking to another constructor, about the same complexity/lines-of-code)

Also, AFAIK this should have no performance impact (tuples and namedtuples are static and have lots of compiler opts.

What do you think?

src/fittable_laws.jl
```diff
@@ -77,3 +77,172 @@ function (law::FM90)(wave::T) where T
     return exvebv
 end
 
+
+"""
+    P92(BKG_amp=218.57,
```
mileslucas, line 82: You'll want to show that these are keyword arguments with the semicolon
```julia
    P92(; BKG_amp=218.57,
```
(and propagate the spacing).

mileslucas (approved):
I have some comments about whether we should restructure the P92 struct, so if we go with that there'll be some changes, but otherwise this looks good to go if you bump the minor version and fix that docstring.

src/fittable_laws.jl
siddharthlal25: Your point is correct, that would make the struct quite organized, but let's think from the user's perspective who is trying to switch from DustExtinction (python) to this place, things like this could be intimidating to the user. What do you think?

mileslucas: We can have both. The user isn't necessarily going to interact with the struct directlyy, just with the _constructors_ and the _methods_. We can organize the struct however we want and organize the constructors however we want. It may be easier to not worry about it

src/fittable_laws.jl
siddharthlal25: Sorry, I clicked on resolved by mistake. 😅

icweaver:
Will be back for you ='(

icweaver (changes requested):
Thanks for picking this up, Abhro! I've just left my minor formatting suggested edits + one science question

src/fittable_laws.jl
```diff
@@ -80,3 +80,161 @@ function (law::FM90)(wave::T) where T <: Real
...
+    axav += _p92_single_term(lam, law.FIR_amp, law.FIR_lambda, law.FIR_b, law.FIR_n)
+
+    return axav
+end
+
+# function for calculating a single P92 term
+function _p92_single_term(x::Real, amplitude::Real, cen_wave::Real, b::Real, n::Real)
+    l_norm = x / cen_wave
+    return amplitude / (l_norm^n + inv(l_norm^n) + b)
+end
```
cgarling:
```suggestion
function _p92_single_term(x::Real, amplitude::Real, cen_wave::Real, b::Real, n::Real)
    l_norm = x / cen_wave
    l_norm_n = l_norm^n
    return amplitude / (l_norm_n + inv(l_norm_n) + b)
end
```
Computing `l_norm^n` once and reusing it improves performance

src/fittable_laws.jl
```diff
@@ -80,3 +80,162 @@ function (law::FM90)(wave::T) where T <: Real
...
+## References
+[Pei (1992)](https://ui.adsabs.harvard.edu/abs/1992ApJ...395..130P)
+"""
+Base.@kwdef struct P92{T<:Number} <: ExtinctionLaw
+    BKG_amp::T = 165.0 * (1 / 3.08 + 1)
+    BKG_lambda::T = 0.047
+    BKG_b::T = 90.0
+    BKG_n::T = 2.0
+    FUV_amp::T = 14.0 * (1 / 3.08 + 1)
+    FUV_lambda::T = 0.07
```
icweaver:
```suggestion
    FUV_lambda::T = 0.08
```

icweaver (changes requested):
Just following up from https://github.com/JuliaAstro/DustExtinction.jl/pull/29#discussion_r2641185813, tests look to be passing with the value from the paper. I tagged this in this current round of reviews. Otherwise this lgtm, thanks again for wrapping up this longstanding PR!

icweaver (approved):

icweaver:
> I'll also leave a ping in that repo to follow up

Update: The value has been changed back to 0.08 in https://github.com/karllark/dust_extinction/pull/266 as well, so I think we're all good here

PR #30, opened 2020-07-07
Added F99 model

icweaver:
- [x] Add support for Measurements
- [x] Generalize for F04

Contributing to #10


giordano:
> Everything looks to be working alright except for Measurements handling. I think this might be related to the package (Dierckx.jl) I am using for interpolation?

I'm not at the computer right now, but if I remember correctly Dierckx.jl is based on an external C library, right? If so, yeah, Measurements.jl can't propagate the uncertainty through `ccall`s, but you can use the `@uncertain` macro: https://juliaphysics.github.io/Measurements.jl/stable/examples/#@uncertain-Macro-1

src/color_laws.jl
```diff
@@ -228,3 +230,143 @@ function gcc09_invum(x::Real, Rv::Real)
...
+
+"""
+    DustExtinction.f99_invum(x, Rv)
+
+The algorithm used for the [`F99`](@ref) extinction law, given inverse microns and Rv. For more information, seek the original paper.
+"""
+function f99_invum(x::Real, Rv::Real)
+    if !(0.3 <= x <= 10.0)
+        error("out of bounds of F99, support is over $(bounds(F99)) angstrom")
+    end
```
siddharthlal25, line 323: I think this is redundant because we already have checked bounds in line number-308.

src/color_laws.jl
```diff
@@ -228,3 +230,143 @@ function gcc09_invum(x::Real, Rv::Real)
...
+        # spline points
+        x_splineval_optir = optnir_axav_x
+
+        # determine optical/IR values at spline points
+        y_splineval_optir = optnir_axav_y
+
+        spline_x = vcat(x_splineval_optir, x_splineval_uv)
+        spline_y = vcat(y_splineval_optir, y_splineval_uv)
+        spl = Spline1D(spline_x, spline_y, k=3)
+        axav = spl(x)
```
siddharthlal25, line 288: I would suggest packing this whole part in an if-else block.

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.7.0"
 
 [deps]
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
+Dierckx = "39dd38d3-220a-591b-8e3c-4c3a8c710a94"
```
giordano, line 8: Note that you should also add compatibility bounds for the package

src/color_laws.jl
icweaver: Roger! Just added in 1901d928043874a9610e6786b9aa2c05ab55bcfa

test/color_laws.jl
```diff
@@ -248,3 +249,40 @@ end
...
+        bad_xs = [0.2, 11.0]
+        @test @inferred(map(law, bad_xs)) == zeros(length(bad_xs))
+        @test_throws ErrorException f99_invum(bad_xs[1], rv)
+        @test_throws ErrorException f99_invum(bad_xs[2], rv)
+
+        # uncertainties
+        #noise = randn(length(wave)) .* 0.01
+        #wave_unc = wave .± noise
+        #reddening = @inferred map(law, wave_unc)
+        #@test Measurements.value.(reddening) ≈ ref_values[rv] rtol = 0.016
```
giordano, line 280: It's weird to use `randn` for the noise, as you can get negative errors, which is not great (`Measurements.jl` doesn't enforce non-negative errors), unless you take the absolute value.  With this change you should be able to test propagation of the uncertainty through `Dierckx.jl`
```suggestion
        noise = rand(length(wave)) .* 0.01
        wave_unc = wave .± noise
        reddening = map(w -> @uncertain(law(w)), wave_unc)
        @test Measurements.value.(reddening) ≈ ref_values[rv] rtol = 1e-3
```

src/color_laws.jl
icweaver: I was wondering about that too. Removing this check seems to cause the following [tests](https://github.com/icweaver/DustExtinction.jl/blob/F99/test/color_laws.jl#L271-L274) to fail (copied below):

```
bad_xs = [0.2, 11.0]
@test @inferred(map(law, bad_xs)) == zeros(length(bad_xs))
@test_throws ErrorException f99_invum(bad_xs[1], rv)
@test_throws ErrorException f99_invum(bad_xs[2], rv)
```

I think the bounds check on line number-308 is only for the top level interface (e.g. `F99(<Rv value>)(<wavelength in Å>`)), while this test is for the lower level input (e.g. `f99_invum(<wavenumber in 1/microns>, <Rv value>`))? 

Definitely agree though that this seems a little redundant.

icweaver:
> > Everything looks to be working alright except for Measurements handling. I think this might be related to the package (Dierckx.jl) I am using for interpolation?
> 
> I'm not at the computer right now, but if I remember correctly Dierckx.jl is based on an external C library, right? If so, yeah, Measurements.jl can't propagate the uncertainty through `ccall`s, but you can use the `@uncertain` macro: https://juliaphysics.github.io/Measurements.jl/stable/examples/#@uncertain-Macro-1

This is so cool! `@uncertain` is working like a charm for single wavelengths. For example, I can do this:
```julia
>>> @uncertain F99()(wave_unc[1])
0.08548387096774193 ± 0.0
```
but I think I am having trouble broadcasting this when I try:
```julia
>>> @uncertain F99().([wave_unc[1], wave_unc[2]])
ERROR: MethodError: no method matching measurement(::Tuple{Array{Measurement{Float64},1}})
```

for example.

giordano:
> but I think I am having trouble broadcasting this when I try:

See https://github.com/JuliaAstro/DustExtinction.jl/pull/30#discussion_r450747097 :wink:

icweaver:
> > but I think I am having trouble broadcasting this when I try:
> 
> See [#30 (comment)](https://github.com/JuliaAstro/DustExtinction.jl/pull/30#discussion_r450747097) 😉

hahaha, was just about to respond. Thanks!

test/color_laws.jl
icweaver: Awesome, thanks! Should this change also be made for the other models' tests?

giordano: If by "this change" you refer to `randn` -> `rand` for the noise, I'd say probably yes.

icweaver: Yep, that was all! Ok, would switching everything over to `rand` be better as a separate PR?

giordano: > would switching everything over to rand be better as a separate PR?

Yes, I think it's better to keep logically separate changes in different pull requests :wink:

Project.toml
icweaver: TIL about compatibility bounds. It looks like versions 0.4.1 and 0.4.2 support Julia 1.0, but I was wondering if there was a hard and fast rule for how far back to go?

giordano: > TIL about compatibility bounds.

A `Project.toml` with compat bounds and a `Manifests.toml` will become your best friends when you want to prepare a script for your analysis pipeline.  You can read more in https://julialang.github.io/Pkg.jl/v1/environments/.

> It looks like versions 0.4.1 and 0.4.2 support Julia 1.0, but I was wondering if there was a hard and fast rule for how far back to go?

You can simply use
```toml
Dierckx = "0.4"
```
see https://julialang.github.io/Pkg.jl/v1/compatibility/

icweaver: Ok, should be up now! cbd9d404e3ec9be3762969c82660be578a2d94e0

Thanks for the additional info. It's definitely been a shift coming from conda land.

giordano: > It's definitely been a shift coming from conda land.

The idea is not much different, you'll create an environment also in this case.  It's just different how you do that.  And the environment is local to Julia, it doesn't overtake your shell, which caused me troubles a few times when trying to run unrelated system programs.

src/color_laws.jl
siddharthlal25: Do astronomers use `invum` function directly? This is not exposed directly to the user in our package. Thoughts? @mileslucas @giordano

src/color_laws.jl
```diff
@@ -228,3 +230,141 @@ function gcc09_invum(x::Real, Rv::Real)
...
+end
+
+"""
+    F99(;Rv=3.1)
+
+Fitzpatrick (1999) dust law.
+
+# References
+[Fitzpatrick (1999)](https://ui.adsabs.harvard.edu/abs/1999PASP..111...63F/)
+"""
```
siddharthlal25, line 300: How about a line describing more about the model (e.g. where is the model applicable)? Check the documentation of other models for reference.

icweaver: Ok, should be up. f458a34791e3c16be89d7c6146c5259988fe80da

src/color_laws.jl
giordano: Yes, this is redundant, good catch!

mileslucas: I see the code as two parts: the public facing API meant for consumption, as well as the internal implementations which are meant to more rigorously represent the original mathematical derivations. This is why the public API returns 0 extinction instead of erroring outside the limits: because nothing makes ~me~astronomers grouchier than having to precisely mask your spectra to match rough empirical limits. However, the `_invum` functions pretty directly translate to the mathematical definitions and I want them to be explicit in their behaviors around the bounds.

As for the bound discrepancies- I would just choose one and keep it consistent between this and `bounds`.

icweaver: Ok, just dropped! d81b97fd8b72bb570ee44f91be64bbfe4e6ee091

I guess these should also be removed from the other models in another PR?

src/color_laws.jl
```diff
@@ -228,3 +230,143 @@ function gcc09_invum(x::Real, Rv::Real)
...
+    gamma = 0.99
+
+    # terms depending on Rv
+    c2 = -0.824 + 4.717 / Rv
+    # original F99 c1-c2 correlation
+    c1 = 2.030 - 3.007 * c2
+
+    # spline points
+    optnir_axav_x = 10000.0 ./
+        [26500.0, 12200.0, 6000.0, 5470.0, 4670.0, 4110.0]
```
mileslucas, line 338: Similar to others- make const if possible and use tuples to avoid allocations
```julia
const optnir_axav_x = aa_to_invum.((26500, 12200, 6000, 5470, 4670, 4110))
```

src/color_laws.jl
```diff
@@ -228,3 +230,143 @@ function gcc09_invum(x::Real, Rv::Real)
...
+    #    Also, fm_unred.pro has different coeff and # of terms,
+    #    but later work does not include these terms
+    #    --> check with Fitzpatrick?
+    opt_axebv_y = [
+        -0.426 + 1.0044 * Rv,
+        -0.050 + 1.0016 * Rv,
+         0.701 + 1.0016 * Rv,
+         1.208 + 1.0032 * Rv - 0.00033 * (Rv^2),
+    ]
+    nir_axebv_y = [0.265, 0.829] * Rv / 3.1
```
mileslucas, line 357: you'll want to make use of broadcasting here-

`nir_axebv_y = @. (0.264, 0.829) * Rv / 3.1`

src/color_laws.jl
icweaver: Added a [post](https://github.com/JuliaAstro/DustExtinction.jl/issues/10#issuecomment-655706169) in the original issue to help keep track of things

icweaver:
Thanks so much for these comments, @mileslucas! I was trying to strike a balance between doing a 1-1 translation from the original astropy code and making things more Julian. Your suggestions have helped answer a lot of "best practices" questions I've had when starting with this language and I'll start implementing them ASAP

src/color_laws.jl
```diff
@@ -228,3 +230,147 @@ function gcc09_invum(x::Real, Rv::Real)
...
+    optnir_axav_x::Vector{<:Real},
+    optnir_axav_y::Vector{<:Real},
+    )
+
+    # x value above which FM90 parametrization used
+    x_cutval_uv = 10000.0 / 2700.0
+
+    # required UV points for spline interpolation
+    x_splineval_uv = 10000.0 ./ [2700.0, 2600.0]
+
```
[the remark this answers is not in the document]
icweaver: 2f4eceb

src/color_laws.jl
```diff
@@ -228,3 +230,147 @@ function gcc09_invum(x::Real, Rv::Real)
...
+    x::Real,
+    Rv::Real,
+    c1::Real,
+    c2::Real,
+    c3::Real,
+    c4::Real,
+    x0::Real,
+    gamma::Real,
+    optnir_axav_x::Vector{<:Real},
+    optnir_axav_y::Vector{<:Real},
```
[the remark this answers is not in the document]
icweaver: 2f4eceb

src/color_laws.jl
```diff
@@ -228,3 +230,147 @@ function gcc09_invum(x::Real, Rv::Real)
...
+    if x >= x_cutval_uv
+        xuv = vcat(x_splineval_uv, x)
+    else
+        xuv = x_splineval_uv
+    end
+
+    # FM90 model and values
+    fm90_model = FM90(c1=c1, c2=c2, c3=c3, c4=c4, x0=x0, gamma=gamma)
+    # evaluate model and get results in A(x)/A(V)
+    axav_fm90 = fm90_model.(10000.0 ./ xuv) / Rv .+ 1.0 # Expects xuv in Å
```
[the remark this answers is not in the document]
icweaver: 2f4eceb

src/color_laws.jl
icweaver: 2f4eceb

src/color_laws.jl
icweaver: 2f4eceb

icweaver:
Ok, I think that just about does it except for the bounds check. I put back the checks and changed it to be consistent with the other tests (29ff98d), except for one part, which seems to be failing. 

For some reason, using `broadcast` instead of `map` gives the following error:
```
F99: Error During Test at /Volumes/GoogleDrive/My Drive/DustExtinction.jl/test/color_laws.jl:253
Got exception outside of a @test
return type Array{Float64,1} does not match inferred return type Union{BitArray{1}, Array}
Stacktrace:
[1] error(::String) at /Applications/Julia-1.4.app/Contents/Resources/julia/lib/julia/sys.dylib:?
[2] top-level scope at /Volumes/GoogleDrive/My Drive/DustExtinction.jl/test/color_laws.jl:268
[3] top-level scope at /Applications/Julia-1.4.app/Contents/Resources/julia/share/julia/stdlib/v1.4/Test/src/Test.jl:1113
[4] top-level scope at /Volumes/GoogleDrive/My Drive/DustExtinction.jl/test/color_laws.jl:256
[5] include(::String) at /Applications/Julia-1.4.app/Contents/Resources/julia/lib/julia/sys.dylib:?
[6] top-level scope at /Volumes/GoogleDrive/My Drive/DustExtinction.jl/test/runtests.jl:8
[7] include(::String) at /Applications/Julia-1.4.app/Contents/Resources/julia/lib/julia/sys.dylib:?
[8] top-level scope at none:6
[9] eval(::Module, ::Any) at /Applications/Julia-1.4.app/Contents/Resources/julia/lib/julia/sys.dylib:?
[10] exec_options(::Base.JLOptions) at /Applications/Julia-1.4.app/Contents/Resources/julia/lib/julia/sys.dylib:?
[11] _start() at /Applications/Julia-1.4.app/Contents/Resources/julia/lib/julia/sys.dylib:?

```

I am not sure where a `Union{BitArray{1}, Array}` is sneaking in when using `broadcast`. This is an issue that I was running into before that I forgot to mention. Does anyone have any ideas?

mileslucas:
I pulled your branch and took a look- I wouldn't worry too much about it, most importantly `@code_warntype F99()(3000)`, for example, is clean- the broadcasting mechanics are beyond the scope of this PR/my worries.

icweaver:
Ok, I'll stick with the current commit that uses `map` then!

icweaver:
Sorry, not sure why this latest commit failed on some platforms while working on others. The local tests I ran before committing all passed.

icweaver:
> Sorry, not sure why this latest commit failed on some platforms while working on others. The local tests I ran before committing all passed.

Ahh, didn't realize `evalpoly` is only for 1.4

https://docs.julialang.org/en/v1/base/math/#Base.Math.evalpoly

src/color_laws.jl
```diff
@@ -228,3 +248,131 @@ function gcc09_invum(x::Real, Rv::Real)
...
+
+The algorithm used for the [`F99`](@ref) extinction law, given inverse microns and Rv. For more information, seek the original paper.
+"""
+function f99_invum(x::Real, Rv::Real)
+    if !(0.3 <= x <= 10.0)
+        error("out of bounds of F99, support is over $(bounds(F99)) angstrom")
+    end
+
+    # terms depending on Rv
+    c2 = @evalpoly (1. / Rv) f99_Rv_params...
```
giordano, line 342: This cannot work with this macro, you need to list the elements of the vector one by one. That's why the function was introduced

icweaver:
Ah, I see now it's failing because you can't splat a Tuple into @evalpoly in Julia v1.0.5

src/color_laws.jl
icweaver: Oh, just saw your comment! Yep, just realized that when I looked at Travis

icweaver: Alrighty, everything should be switched over now. I've never had to think about things like backwards compatibility before, so this has been an extremely useful lesson for me, thanks!

siddharthlal25:
> According to the discussion [here](https://discourse.julialang.org/t/cubic-interpolation/34552), it looks like Interpolations.jl does not have support for cubic B-splines (which is used by the python [implementation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.splrep.html)), while this package does.

I was just going through `Interpolations.jl` and found [this](http://juliamath.github.io/Interpolations.jl/stable/control/#BSplines-1), is this what you were looking for?

icweaver:
> > According to the discussion [here](https://discourse.julialang.org/t/cubic-interpolation/34552), it looks like Interpolations.jl does not have support for cubic B-splines (which is used by the python [implementation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.splrep.html)), while this package does.
> 
> I was just going through `Interpolations.jl` and found [this](http://juliamath.github.io/Interpolations.jl/stable/control/#BSplines-1), is this what you were looking for?

Right, so my understanding was the issue was with the [next part down](http://juliamath.github.io/Interpolations.jl/stable/control/#Scaled-BSplines-1), which I believe @mileslucas mentions [here](https://github.com/JuliaAstro/DustExtinction.jl/pull/30#pullrequestreview-445028416), re: cubic splines on a non-uniform grid?

src/color_laws.jl
icweaver: Just closing the loop on some of our threads. The test as been put back re: https://github.com/JuliaAstro/DustExtinction.jl/pull/30#discussion_r451765431

icweaver:
Following up on the spline issue, there is this [package](https://github.com/sp94/CubicSplines.jl) that could be useful!

* Pros: It's written in pure Julia (which should make it play nice with AD?) and can do cubic splines on an irregular grid
* Cons?: Those cubic splines are actually Akima splines, so our results will differ from astropy implementation of the dust models

I have no experience using Akima splines, but it looks like it may actually [perform better than traditional cubic B-splines](http://www.jive.nl/jivewiki/lib/exe/fetch.php?media=expres:fabric:interpolation.pdf).

I guess it comes down to: should DustExtinction.jl exactly reproduce its Python cousin's results, or should it use the "latest and greatest" technology?

src/color_laws.jl
```diff
@@ -10,6 +12,22 @@ const ccm89_cb = [0.0, 1.41338, 2.28305, 1.07233, -5.38434, -0.62251, 5.3026, -2
...
+# spline points
+const f99_optnir_axav_x = aa_to_invum.((26500, 12200, 6000, 5470, 4670, 4110))
+const f99_nir_axebv_y_params = @. (0.265, 0.829) / 3.1
+
+# c1-c2 correlation terms
+const f99_c3 = 3.23
+const f99_c4 = 0.41
+const f99_x0 = 4.596
+const f99_gamma = 0.99
+
```
mileslucas, line 30: I think these would fit in better closer to the F99 definition- you'll see things are generally clustered by law already

icweaver: For sure 804fb33

icweaver:
It looks like the CI sporadically fails when installing Julia, but otherwise I think everything should be addressed now for another review. Thanks!

src/color_laws.jl
```diff
@@ -228,3 +230,146 @@ function gcc09_invum(x::Real, Rv::Real)
...
+    # ignore the spline points
+    if x >= f99_x_cutval_uv
+        axav = last(axav_fm90)
+    else
+        # **Optical Portion**
+        # using cubic spline anchored in UV, optical, and IR
+
+        # optical/NIR points in input x
+
```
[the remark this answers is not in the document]
icweaver: Oh, whoops! That's some [leftover Python styling](https://github.com/karllark/dust_extinction/blob/master/dust_extinction/shapes.py#L114-L118) before I switched to thinking in terms of "kernels". Thanks for catching that, should be fixed now! (5a6b6fd)

mileslucas:
Actually, if you want to go ahead and edit `Project.toml` and bump the version to 0.8.0 we'll be prepped for a release.

icweaver:
> This is looking good! Thanks again for the contribution. The only changes left are completely internal and I'm not too concerned. Unless anybody else has concerns I'll merge this later tonight and do a new release.

Awesome! I opened up a separate PR for F04 (#31) and F19 (#32) as well. I tried to keep in mind everything y'all have taught me =]

PR #31, opened 2020-07-14
Added F04 model

icweaver:
https://dust-extinction.readthedocs.io/en/stable/api/dust_extinction.parameter_averages.F04.html?highlight=F04

Contributing to #10


giordano:
> so the above is just my workaround to make F04 compatible in all versions.

Could you please leave a comment in the code so that when we'll drop support for Julia v1.0 we can remove this workaround?

icweaver:
Sure, just added! 1f6fae8

haha, it's kind of wild to think about leaving 1.0 behind some day ='(

giordano:
> haha, it's kind of wild to think about leaving 1.0 behind some day ='(

Not really, there is the general hope of making Julia v1.6 the new Long Term Support version (but the decision will be taken after its release), at which point Julia v1.0 will not be supported anymore :wink:

PR #32, opened 2020-07-14
Added F19 model

icweaver:
https://dust-extinction.readthedocs.io/en/stable/api/dust_extinction.parameter_averages.F19.html?highlight=F19

Contributing to #10


giordano:
You've got quite some merge conflicts to fix :slightly_smiling_face:

icweaver:
> You've got quite some merge conflicts to fix 🙂

Oh, yep. I was debating if the parallel branches I was making should talk to each other or not. I went with the silent treatment and merge manually route because I wasn't sure if F99 would be merged first or not. Glad it was!

src/color_laws.jl
```diff
@@ -370,3 +370,63 @@ function f99_invum(x::Real, Rv::Real)
             optnir_axebv_y,
         )
 end
+
+# read and unpack tabulated data
+fpath = joinpath(dirname(pathof(DustExtinction)), "data/F19_tabulated.dat")
+data = readdlm(fpath, skipstart=1)
+const data_x, data_k, data_delta_k, data_sigma_k = (data[:, i] for i in 1:4)
```
giordano, line 377: Defining `fpath` in this way will not work on Windows.  Also, you can avoid polluting the namespace with variables that you are not going to later use with a [`let`](https://docs.julialang.org/en/v1/base/base/#let):
```suggestion
const data_x, data_k, data_delta_k, data_sigma_k = let data = readdlm(joinpath(dirname(pathof(DustExtinction)), "data", "F19_tabulated.dat"), skipstart=1)
    (data[:, i] for i in 1:4)
end
```

test/color_laws.jl
```diff
@@ -286,3 +287,41 @@ end
...
+
+        bad_waves = [100, 4e4]
+        @test @inferred(map(law, bad_waves)) == zeros(length(bad_waves))
+        @test_throws ErrorException f19_invum(aa_to_invum(bad_waves[1]), rv)
+        @test_throws ErrorException f19_invum(aa_to_invum(bad_waves[2]), rv)
+
+        # uncertainties
+        noise = rand(length(wave)) .* 0.01
+        wave_unc = wave .± noise
+        reddening = map(w -> @uncertain(law(w)), wave_unc)
```
giordano, line 318: Do you need to use `@uncertain` also here?

icweaver: I think so, since this model calls Dierckx.jl's wrapping of the Fortran library that's doing the interpolation, like in F99 I believe.

giordano: Ok, I asked for confirmation :slightly_smiling_face:

src/color_laws.jl
icweaver: Oh, really cool! I'll read up more on `let`

test/color_laws.jl
icweaver: I just realized it would have been perfect if Fitzpatrick et al. created interpolated models in 1900/1995. The we would have F90/F95 models calling F90/F95!

src/color_laws.jl
icweaver: Also, completely forgot about the forward slash I left behind, thanks!

src/color_laws.jl
```diff
@@ -372,9 +372,9 @@ function f99_invum(x::Real, Rv::Real)
 end
 
 # read and unpack tabulated data
-fpath = joinpath(dirname(pathof(DustExtinction)), "data/F19_tabulated.dat")
```
icweaver: Sorry, think my question was lost in the merge. Should we be worried that AppVeyor didn't seem to catch the forward slash I accidentally left behind?

src/color_laws.jl
```diff
@@ -370,3 +370,63 @@ function f99_invum(x::Real, Rv::Real)
             optnir_axebv_y,
         )
 end
+
+# read and unpack tabulated data
+const data_x, data_k, data_delta_k, data_sigma_k = let data = readdlm(joinpath(dirname(pathof(DustExtinction)), "data", "F19_tabulated.dat"), skipstart=1)
```
icweaver, line 375: Sorry, think my question was lost in the merge. Should we be worried that AppVeyor didn't seem to catch the forward slash I accidentally left behind?

giordano: No, but I'm surprised it didn't complain.

src/data/F19_tabulated.dat
```diff
@@ -0,0 +1,103 @@
+x      k_3.02  deltak  sigmak
```
mileslucas, line 1: we are actually already making use of a data-providing package `DataDeps`. If this data is available online already, I suggest we switch to using DataDeps for providing it.

test/color_laws.jl
mileslucas: Ah yes, Fitzpatrick '00, making use of the many mid-infrared space telescopes we had in 1900 😁

src/data/F19_tabulated.dat
icweaver: Oh yea, I saw that was used for the dust map. For sure!

test/color_laws.jl
icweaver: lol, just edited

src/data/F19_tabulated.dat
icweaver: Ok, just tried it out (fef8bdc)

At first I tried sticking with the using constants approach by just directly swapping out the read path from this:
```julia
const data_x, data_k, data_delta_k, data_sigma_k = let data = readdlm(joinpath(dirname(pathof(DustExtinction)), "data", "F19_tabulated.dat"), skipstart=1)
    (data[:, i] for i in 1:4)
end
```

to this:
```julia
const data_x, data_k, data_delta_k, data_sigma_k = let data = readdlm(joinpath(datadep"F19", "F19_tabulated.dat"), skipstart=1)
    (data[:, i] for i in 1:4)
end
```

where `F19` is the data directory I registered in `src/DustExtinction.jl`. This seems to work fine if I just run parts of the package (i.e. importing DataDeps, then running just the lines above), but attempting import the whole package leads to a KeyError when trying to point to the data via `datadep"F19"`. 

For some reason, it looks like the string macro can't be resolved unless it is inside a function scope? I tested this out with the same error for `datadep"sfd98_map"` by just trying to print it right above [where it is called in a function](https://github.com/JuliaAstro/DustExtinction.jl/blob/master/src/dust_maps.jl#L58), so I am pretty confused about why this doesn't work.

For now I just put everything inside the internal `f19_invum` function (fef8bdc), but I don't know how that compares in terms of performance to just using constants.

icweaver:
> I think a few changes from #31 slipped into this pull request

Right, this goes back to my [internal debate](https://github.com/JuliaAstro/DustExtinction.jl/pull/32#issuecomment-658520994) about manually merging parallel branches. Since F04 has already been approved, I figured I'd get a jump on merging my F04 branch into this one before this branch of my fork is pulled into the main repo.

Since I'm editing the same lines of the same file in my different branches, I guess a merge conflict is inevitable either way and I'm really just doubling the work for myself by doing it this way. I guess merging model branches together in a PR also pollutes them and just makes more work for reviewers to read through duplicate changes. I'll roll back my merge commit and stick to my strategy I was using before (waiting for another branch that edits the same lines to be merged to master before manually merging the changes)!

mileslucas:
Please rebase on master, and from here bump to 0.10.0 when ready. I'm realizing while I write this that I probably should've just bundled these, but too late.

icweaver:
Ok, looks like all tests are passing, everything has been merg-ified, and version has been bumped 👍🏾

icweaver:
> Please rebase on master, and from here bump to 0.10.0 when ready. I'm realizing while I write this that I probably should've just bundled these, but too late.

Just checking, would you also like for some commits to be squashed?

icweaver:
haha, for sure

giordano:
> I've never cared about commit squashing.

However note that having as few commits as possible makes using tools like `git blame` and `git bisect` _way_ easier. The best would be to have few, logically consistent, and working revisions.

PR #33, opened 2020-07-16
Added M14 model

icweaver:
https://dust-extinction.readthedocs.io/en/stable/api/dust_extinction.parameter_averages.M14.html?highlight=M14

Contributing to #10


icweaver:
> 5cdb724

Seems to be failing to build `Dierckx.jl` in Julia 1.0 on FreeBSD and Julia 1.4 in Linux Xenial?

src/color_laws.jl
```diff
@@ -452,3 +452,110 @@ function f04_invum(x::Real, Rv::Real)
...
+    # Infrared
+    ai = 0.574 * x^1.61
+    bi = -0.527 * x^1.61
+
+    # Optical
+    m14_av = m14_a_spl(x)
+    m14_bv = m14_b_spl(x)
+
+    a = ai * (x < m14_xi1) + m14_av * ((x >= m14_xi1) & (x < m14_xi3))
+    b = bi * (x < m14_xi1) + m14_bv * ((x >= m14_xi1) & (x < m14_xi3))
```
giordano, line 558: I know this is perfectly legal and very efficient, but I find these lines hard to read, it took me a while to realise that this should be equivalent to something like
```julia
a = zero(m14_av)
b = zero(m14_bv)
if x < m14_xi1
    # Infrared
    a = 0.574 * x^1.61
    b = -0.527 * x^1.61
elseif x < m14_xi3
    # Optical
    a = m14_a_spl(x)
    b = m14_b_spl(x)
end
```
which also saves you some computations above.  An alternative would be
```julia
a = x < m14_xi1 ? ai : x < m14_xi3 ? m14_av : zero(m14_av)
b = x < m14_xi1 ? bi : x < m14_xi3 ? m14_bv : zero(m14_bv)
```

icweaver: Wow, that reads so much better, thanks! Just changed it c042984

giordano:
A comment about version numbers: new features in the 0.x series don't require a minor version bump.  Unless we break the API, we can simply bump the patch version, avoiding releasing a breaking version each time, creating annoying headaches to anyone that cares about version compatibilities.

icweaver:
> Seems like some unneccessary comments can be removed, otherwise rebase and I can release as part of v0.10

Ok, should be good to go

PR #34, opened 2020-07-17
Added G16 model

icweaver:
https://dust-extinction.readthedocs.io/en/latest/api/dust_extinction.parameter_averages.G16.html#dust_extinction.parameter_averages.G16

Contributing to #10 

I split this one out into a separate "mixture models" section. Happy to move things around as needed


src/mixture_laws.jl
```diff
@@ -0,0 +1,130 @@
...
+end
+
+"""
+    G16(;Rv=3.1)
+
+Gordon et al. (2016) Milky Way, LMC, & SMC R(V) and f_A dependent model
+
+Returns E(B-V) in magnitudes at the given wavelength relative to the
+extinction. This is mixture model between the F99 R(V) dependent model
+(component A) and the G03_SMCBar model (component B) The default support is
```
mileslucas, line 86:
```suggestion
(component A) and the [`G03_SMCBar`](@ref) model (component B) The default support is
```

you accidentally italicized the rest of the docstring

src/mixture_laws.jl
```diff
@@ -0,0 +1,130 @@
...
+[1000, 33333] Å. Outside of that range this will return 0. Rv is the selective
+extinction and is valid over [2, 6]. A typical value for the Milky Way is 3.1.
+By default, Rv=3.1 and f_A=1.0 is used.
```
mileslucas, line 89:
```suggestion
```
Defaults are in the doc string

src/mixture_laws.jl
```diff
@@ -0,0 +1,130 @@
...
+        g03_C4,
+        g03_x0,
+        g03_gamma,
+        g03_optnir_axav_x,
+        g03_optnir_axav_y,
+    )
+end
+
+"""
+    G16(;Rv=3.1)
```
mileslucas, line 80:
```suggestion
    G16(;Rv=3.1, f_A=1.0)
```
don't lie ;)

icweaver: ahh, you caught me

icweaver:
> Looks pretty good! Looking at the docs I've realized how much cooler having sliders with plotly plots would be.

Interact with all the things

src/mixture_laws.jl
```diff
@@ -95,7 +95,7 @@ extinction and is valid over [2, 6]. A typical value for the Milky Way is 3.1.
     f_A = 1.0
 end
 
-G16(Rv, f_A) = G16(promote(Rv, f_A)...)
+#G16(Rv, f_A) = G16(promote(Rv, f_A)...)
 #G16(Rv=3.1, f_A=3.1) = G16(Rv, f_A)
```
icweaver, line 99: I was also curious about what this line does. Is it a multiple dispatch thing?

mileslucas: `promote` is a complex mechanism but it ensures that the outputs are all of the same type. To get a rough idea, look at the following examples
```julia
promote(1, 1.0)
promote(2.0, 3.5)
promote(3f0, 1)
promote(3f0, 1e2)
promote("1", 1)
```

it finds the "lowest common denominator" type for all inputs. A common design pattern is to type-def a struct such as 
```julia
struct TrivialStruct{T}
	a::T
	b::T
end
```
which means both fields `a` and `b` are of type `T` (look up parametric structs for more info). This means if I call `Trivial
Struct(1, 1.0)` it fails, because `1` and `1.0` do not have the same type. If I call `TrivialStruct(promote(1, 1.0)...)` it will promote the two inputs to the closest common type and use that as the input for the struct. None of this is required: it's just common practice for parametric structs!

icweaver: Right, thanks Miles. I appreciate the examples! I just realized I the section I highlighted could look misleading. My question was more about the uncommented bit directly below.

```julia
G16(Rv=3.1, f_A=3.1) = G16(Rv, f_A)
```

Unless that is what you meant and it is also a promotion thing? Sorry, still getting used to the new syntax!

mileslucas: Oh sorry I misunderstood. That line is a function definition. We're using it to set defaults for a constructor. We do this because the with_kw macro makes a keyword argument only method.

mileslucas:
Thanks for your patience in getting this merged.

src/mixture_laws.jl
icweaver: Oh, that's pretty neat. Thank!

icweaver:
haha, no rush over here

PR #41, opened 2022-11-10
switch to GHA

mileslucas:
- add github actions workflow
- remove travis and appveyor
- update badges


giordano:
CI was red everywhere?

mileslucas:
ah sorry I didn't see you commented- I didn't include an environment variable, but when I read the logs I mistakenly thought it was because of a compat issue (which was only true for the docs deployment). I had intended to fix compat in #42 so I was going to merge and deal with it separately. It's all good on master now.

PR #47, opened 2024-07-16
Turn off CI testing on nightly

abhro:


abhro:
Better option would be to keep testing and allow failures, but that's a more involved change. Currently testing on nightly clutters up the CI for all other PRs and commits. A future PR can re-add testing on nightly with failures allowed.

PR #50, opened 2025-02-06
LMC Bar method added

stefanocovino:
Dear friends,

I've added, some time ago, the LMC average model in Gordon et al. (2003) to the list of the available recipes. Might it be of interest for the main branch?

  Stefano


test/mixture_laws.jl
```diff
@@ -85,3 +90,38 @@ end
...
+    x, y, tolerance = model.obsdata_x, model.obsdata_axav, model.obsdata_tolerance
+    w = 1e4 ./ x
+    @test model.(collect(w)) ≈ collect(y) rtol = tolerance
+
+    # Test out of bounds
+    model = G03_LMCAve(Rv=2.74)
+    bad_waves = [100, 4e4]
+    @test model.(bad_waves) == zeros(length(bad_waves))
+    @test_throws ErrorException g03_invum(aa_to_invum(bad_waves[1]), 3.1)
+    @test_throws ErrorException g03_invum(aa_to_invum(bad_waves[2]), 3.1)
```
cgarling, line 108: Are you intending to test `g03lmc_invum` here? Currently these tests fail because `g03_invum` throws a `DomainError` while your  `g03lmc_invum` throws a generic `ErrorException`. I prefer the `DomainError` here.
```suggestion
    @test_throws DomainError g03lmc_invum(aa_to_invum(bad_waves[1]), 3.1)
    @test_throws DomainError g03lmc_invum(aa_to_invum(bad_waves[2]), 3.1)
```

src/mixture_laws.jl
```diff
@@ -76,6 +134,29 @@ function g03_invum(x::Real, Rv::Real)
     )
 end
 
+
+function g03lmc_invum(x::Real, Rv::Real)
+    if !(0.3 <= x <= 10.0)
+        error("out of bounds of G03_LMCAve, support is over $(bounds(G03_LMCAve)) angstrom")
```
cgarling, line 140: Recommend making this a `DomainError` for consistency with `g03_invum`
```suggestion
throw(DomainError(x, "out of bounds of G03_LMCAve, support is over $(bounds(G03_LMCAve)) angstrom"))
```

src/mixture_laws.jl
```diff
@@ -15,17 +15,49 @@ const g03_obsdata_axav = (
 
 const g03_obsdata_tolerance = 6e-2
 
+# Data obtained by the DustExtinction python package are slightly different 
+# compared to what it is reported in Table A4 in Gordon et al. (2003)
```
cgarling, line 19: Are these values following DustExtinction or Table A4 in Gordon+2003? It is good to note that they are different but would also be nice to indicate which are being used here.

cgarling (changes requested):
Thanks for the PR! Just a few comments regarding tests and domain errors.

test/mixture_laws.jl
stefanocovino: Actually, I just cloned what I found (when I forked the package) for the  g03smc analogous function, You proposals are of course fine for me.

src/mixture_laws.jl
stefanocovino: Yes, the same as before. Fine for me.

src/mixture_laws.jl
stefanocovino: I've been using data from the python package. I just took note of the difference.

stefanocovino:
I see there were several updates in the main branch I did not include. My fault, of course. As a matter of fact, this fork was ready since several months, but I never manage to pull a request for a long story (I was in contact with one of the developers of this package that, I guess, eventually left the role but I was waiting for comments by him). Anyway, I did not change any part of the code apart from adding the g03lmc recipe.

icweaver:
Thank you for this all! Should some docs and a plot also be added, similar to past models?

stefanocovino:
Hi,

I have thought about this. However, the LMC extinction curve I've added is from a paper (Gordon et al. 2003) already considered in the package. And there are no plots based on these recipes. So, I did not prepare any plot. Nevertheless, if needed, it would be easy to prepare some.

cgarling:
I pushed a commit to add the (short) docstring to the `@docs` block but idk about the plot -- it looks like they are checked into the repo but I usually prefer to generate them on-the-fly during the CI run so that the git history tracks the code that generated the plots rather than the image files themselves. I also find plots overlaying different laws at fixed `R_V` more useful than having one plot per law showing multiple `R_V` values, but that's a personal preference. 

I would be happy to merge as-is and worry about the plot at a later time.

stefanocovino:
I agree.

icweaver:
Alrighty, let's get it y'all. Thanks @stefanocovino for this really nice PR!

cgarling:
Sorry I didn't see your edit -- I don't usually worry about squashing unless I know there's a lot of junk commits that I don't think have value (i.e. I did something wrong and just pushed a new commit rather than reverting).

PR #54, opened 2025-02-13
Migrate away from Dierckx.jl

icweaver:
Addressing the AD concern [brought up by Miles](https://github.com/JuliaAstro/DustExtinction.jl/pull/30#pullrequestreview-445028416)

AFAICT, Interpolations.jl still doesn't seem to have ootb support for cubic B-splines on irregular grids [just yet](https://github.com/JuliaMath/Interpolations.jl/issues/556)

[BSplineKit.jl](https://github.com/jipolanco/BSplineKit.jl) seems like a very nice drop-in replacement though. The output differs slightly from [Dierckx.jl](https://github.com/JuliaMath/Dierckx.jl), but still seems to be within our test tolerances. Might just be a boundary condition thing I'm doing wrong, but still need to explore more

[DataInterpolations.jl](https://github.com/SciML/DataInterpolations.jl) could also be the way to go. Do folks have other thoughts?


cgarling:
I've used DataInterpolations.jl when I had some data that their specialty interpolators were useful for. I don't really like how their syntax to construct an interpolator is `interpolator(y, x)` rather than `(x, y)` but that's a minor complaint. I have no experience with BSplineKit.jl but if it covers our needs (which seem relatively minor) I'm happy to use it. 

Should this be wrapped into the 0.11.2 release? We updated the Project.toml in #50 but we can wait to register until after merging this if you want.

src/color_laws.jl
```diff
@@ -1,4 +1,4 @@
-using Dierckx
+using BSplineKit
```
cgarling, line 1: I would this to be an explicit import (see #55).

cgarling, line 1:
```suggestion
using BSplineKit: interpolate, BSplineOrder, Natural
```

src/color_laws/fitzpatrick.jl
```diff
@@ -262,7 +262,9 @@ and Rv. For more information, seek the original paper.
 """
 function f19_invum(x::Real, Rv::Real)
     # read and unpack tabulated data
-    data_x, data_k, data_delta_k, data_sigma_k = let data = readdlm(joinpath(datadep"F19", "F19_tabulated.dat"), skipstart=1)
+    # using type annotations so `interpolate` can be inferred
+    # TODO: Avoid using readdlm to avoid this issue altogether
+    data_x::Vector{Float64}, data_k::Vector{Float64}, data_delta_k::Vector{Float64}, data_sigma_k::Vector{Float64} = let data = readdlm(joinpath(datadep"F19", "F19_tabulated.dat"), skipstart=1)
         (data[:, i] for i in 1:4)
     end
```
cgarling, line 269: With the caveat that I'm not familiar with this code, it seems to me it is inefficient to re-read the data for every call to `f19_invum`.

Why don't we load it as a constant (something like `const f19_data`) at the module level? I guess that would force download the datadep when you load the package but that sounds fine to me. That would also allow it's type to be known.

Frankly the data file's not even that large so we could just copy-paste the raw data into the source file unless there's some licensing issue or something ...

Project.toml
```diff
@@ -15,7 +15,6 @@ UnitfulAstro = "6112ee07-acf9-5e0f-b108-d242c714bf9f"
 [compat]
 DataDeps = "0.7"
```
cgarling, line 16: Looks like the last version of BSplineKit.jl to support Julia 1.6 (our current bound) was v0.11 -- see [here](https://github.com/jipolanco/BSplineKit.jl/compare/v0.11.0...master). The CI run on Julia 1.6 should let us know if this works or not.

```suggestion
BSplineKit = "0.11"
DataDeps = "0.7"
```

src/color_laws/fitzpatrick.jl
icweaver: 100% agree. I was very tempted to just inline it while writing out that TODO. Will play around with some schemes later tn

Project.toml
icweaver: Thanks for checking this. I have next to no experience with anything versioning related

icweaver:
> I've used DataInterpolations.jl when I had some data that their specialty interpolators were useful for. I don't really like how their syntax to construct an interpolator is `interpolator(y, x)` rather than `(x, y)` but that's a minor complaint. I have no experience with BSplineKit.jl but if it covers our needs (which seem relatively minor) I'm happy to use it.

Awesome, thanks for the real world feedback! I'll try it out tn

> Should this be wrapped into the 0.11.2 release? We updated the Project.toml in #50 but we can wait to register until after merging this if you want.

Sure, thanks! This is shaping up to be a real nice release

giordano:
> I wonder if artifacts might be what's needed to track and work with datasets internally like this?

I haven't checked what's the format of the data you're dealing with, but a note about Artifacts is that it supports only tarballs, if you don't have a tarball to start with you'd need to repackage it and upload it somewhere, which isn't always a nice thing to do (but ArtifactsUtils.jl makes this stuff easier)

cgarling:
It's just a small matrix with a few kb of data (something like 4 columns, 100 rows). Since the data are so small I prefer this solution, and +1 for keeping the link to the raw data file in a comment so we know where it came from. 

lgtm

icweaver:
Thanks all. Yea, that's a good point Mosè about the additional infra that would be needed for an artifacts tarball

Agreed about just using the commented citation for now, Chris. Just fixed a small typo I had in there, but otherwise please feel free to merge if it looks good to you all. Thanks again all for the review!

cgarling:
Oh, I would also remove the datadep registration in the main DustExtinction.jl file for F19 since we are no longer using it.

icweaver:
oop, good catch! Two shakes

PR #58, opened 2025-02-14
Doc plots with Makie

### LINKED ISSUE #36: Plotting recipes (in order to clean up doc plots)
We have all of the information required to plot a dust law in the `ExtinctionLaw` structs, so we should be able to create a plotting recipe with `RecipesBase` to enable usage like
```julia
law = CCM89()
plot(law)
```

creating this plot recipe should be pretty easy and very similar to the [recipes used in Polynomials.jl](https://github.com/JuliaMath/Polynomials.jl/blob/master/src/plots.jl).

In terms of the interface, I think we need three key components
1. what law are we plotting
2. what wavelengths are we plotting
3. Are we plotting wavenumbers (inverse micron, same effective shape as frequency) or wavelengths.

Here's what I think would be easiest to accomplish that interface
1. Trivial
2. By default, use the `bounds` of the law. Add a recipe with bounds `f(::ExtinctionLaw, a, b)` and a recipe with an iterable `f(::ExtincitonLaw, X)`
3. Let's add a keyword argument "unit". Default is `unit="invum"`. Other string options include `unit="angstrom"`, `unit="micron"`, etc. Really we should just use Unitful for parsing these, except for the `"invum"` case since that won't show up in unitful despite being (AFAIK) a de facto unit.

The onus for these recipes is to stop doing what we're doing with the plot docs, which is writing scripts that must be evaluated locally to generate plots that are checked out into version control and then only used for docs. A much better workflow would be to have the plots made as part of the documentation step, and ideally we can show the plotting code inside the docs for full reproducibility. After making these recipes, the docs should be able to be rewritten without the plots dir, and instead with simple `plot(law)` calls using Documenter's `@example` blocks.

icweaver:
Closes #36

Depends on:
- [x] #67: informs the y-axis label units used
- [x] https://github.com/MakieOrg/Makie.jl/pull/4853: Include `Gain` unit support in Makie upstream
- [ ] Tag new release after merge

### Preview
- [New plotting section](https://juliaastro.org/DustExtinction.jl/previews/PR58/plotting/)
- [Color laws](https://juliaastro.org/DustExtinction.jl/previews/PR58/color_laws/)
- [Dust Maps](https://juliaastro.org/DustExtinction.jl/previews/PR58/dust_maps/). I also switched to Cividis here. I think it is a better colormap for accessibility, but can change it back if folks want

### Motivation
DustExtinction.jl currently uses Plots.jl to generate a bunch of static plot images that are then added to the docs. A while back, [Miles had the idea](https://github.com/JuliaAstro/DustExtinction.jl/issues/36) to streamline this process by making the plot creation part of the doc creation process, assisted by plot recipes. Although the original idea was to try and implement this with [Plots.jl's recipe system](https://docs.juliaplots.org/dev/recipes/#recipes), Makie.jl has since come on the scene as a mature, pure Julia alternative.

Noteable examples in JuliaAstro-adjacent ecosystems also exploring and/or already making the switch include [DimensionalData.jl](https://github.com/rafaqz/DimensionalData.jl/issues/815#issuecomment-2393822170) –- which backs the underlying data in AstroImages.jl, [PairPlots.jl](https://github.com/sefffal/PairPlots.jl/pull/12#issue-1513106532), and [SkyImages.jl](https://github.com/JuliaAPlavin/SkyImages.jl/tree/master/ext).

### Major changes
- Replaces Plots.jl with Makie.jl
- Instead of generating plots in a separate script and importing the images into the docs, this generates the plots directly in the docstrings
- This is made more ergonomic with [Makie.jl's recipe system](https://docs.makie.org/v0.22/explanations/recipes)
- Additionally, Julia's new [package extension system](https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) is used to conditionally load the recipes only if the user has Makie.jl installed on their system. [More info here](https://docs.makie.org/v0.22/explanations/recipes#Makie-Package-Extension).

### TODO
- [x] Implement package extension
- [x] Generate parameter average and dust map plots
- [x] Add remaining models to docs
- [x] Cleanup
- [x] Add more text to docs
- [x] ~Bump oldest Julia version to at least v1.9 to use package extensions. Does this sound reasonable?~ https://github.com/JuliaAstro/DustExtinction.jl/pull/58#issuecomment-2662103125
- [x] Tests
    - [x] `Makie.PointLike` (lines, scatter, etc.)
    - [x] `Makie.CellGrid` (heatmaps)


icweaver:
Sweeeet, just saw this too https://github.com/JuliaPhysics/Measurements.jl/pull/174

"Oh yea, it's all coming together.jpeg"

cgarling:
Great start, thanks! 👍 

You intend to export the plotting recipes to users through the package extension? Originally I thought the vision was to define the plotting recipes *just* for the docs so the plotting dependencies could go into the documentation project and not affect the main project.

Typically I find that plotting recipes are useful when you want to make the same plot many times, as we do in the documentation, but I rarely use them otherwise because they cover one use case that usually doesn't match what I want. For example, the plotting recipe shows the reddening law over the full wavelength range with `x` in inverse microns (which is good for in the documentation!), but a situation in which I more commonly find myself (as a user) is wanting to plot the reddening law over a small wavelength range (say, the wavelength coverage of HST/ACS) with `x` in angstroms or nanometers.

This is to say that I don't think I would ever call these plotting recipes as a user myself and so my preference would be to isolate them to the documentation build so that the dependencies of the main code do not need to be changed. This would also allow us to continue iterating on improvements to the docs / plots without worrying how it would affect the main package. This is a weak preference and I'm happy to be persuaded 😄

icweaver:
Thanks for your feedback, I think those are very fair points

What if we restrict the recipe plots (`lplot, mplot, dplot`, and friends) to the docs, but still keep the [building blocks](https://juliaastro.org/DustExtinction.jl/previews/PR58/plotting/#Building-blocks) exposed via the package extension? That way users could still make their own custom plots with Makie's primitives (`lines`, `scatter`, etc.) if they want, like for the HST/ACS wavelength coverage example you mentioned, while keeping our original API unchanged

This would also greatly reduce the scope of the full recipe plots that I would need to consider 😅

cgarling:
The building blocks being the `Makie.convert_arguments` definitions and `Makie.plottype`? That sounds good!

As extensions are optional, something I've done in the past is add extensions while keeping compatibility with older Julia versions (i.e., not bumping julia compat). The extensions will just be ignored if you use the package on an older version, but at least you can still use the latest version of the base package. I think it's nice to keep long back compatibility for foundational packages like this and I'm not sure it's worth jumping from Julia 1.6 -> 1.9 compatibility for a few plotting definitions (are the building blocks for the plotting really just 7 lines? pretty cool). What do you think about this solution?

Alternatively if you really want the makie extension to be available everywhere you could fall back to using Requires.jl on older Julia versions (see, e.g., [here](https://github.com/SciML/DiffEqBase.jl/blob/c2de920a88a39814ac368f703c40fbc282a9c078/src/init.jl)), but I think it's probably not worth the hassle. Actually now that I think about it, using the Requires.jl trick would be worse from a maintenance perspective than just bumping the julia compat so disregard this paragraph

icweaver:
Exactly!

Sounds good. I made the following broad changes:

* Moved the doc-specific plotting functions to their own file: `plotting.jl`
* Call this file from the relevant markdown files via `@setup`
* Run the plotting code from named `@example` blocks, and hide the input via `# hide` comments

I'll keep adding to the documentation, but the overall organization scheme should be ready now for a first pass at reviewing whenever you have the time. I'm still learning the ins-and-outs of Documenter.jl, so if anything looks wonky that was probably a mistake. Thanks!

icweaver:
Also happy to split the doc writing bits out into its own PR if that is preferred 👍🏾

icweaver:
Ok, I think things should be in a pretty good place for a review now. Thanks for all the preliminary comments!

icweaver:
Also, opened #59 to keep track of this comment in case we happen to find more reasons to potentially modernize things

> I think it's nice to keep long back compatibility for foundational packages like this and I'm not sure it's worth jumping from Julia 1.6 -> 1.9 compatibility for a few plotting definitions

icweaver:
Ok, this should be ready for review too now

cgarling (commented):
The [PR preview](https://juliaastro.org/DustExtinction.jl/previews/PR58/) looks good! Thanks for the work. Much prefer this to checking images into the repo. I don't use Makie myself so I can't really comment on the plotting code but it generally looks good to me. Left one comment on perhaps reorganizing the "color laws" page.

cgarling:
I would suggest bumping patch version to v0.11.2 after merging this PR so that the corrected docstrings from #67 and the newly updated plots are reflected in the "stable" documentation. Thanks for your work on these improvements @icweaver!

icweaver:
Thanks, Chris! I think the only thing left is getting https://github.com/MakieOrg/Makie.jl/pull/4853 in. They added it to their [ready to review column](https://github.com/orgs/MakieOrg/projects/12/views/1) last month, so fingers crossed

icweaver:
Whoo! With Makie's PR now in, is there anything else left that we should hit, besides tagging a new release after?

cgarling:
Looks like our changes were incorporated into Makie 0.22.8 (see [here](https://github.com/MakieOrg/Makie.jl/compare/v0.22.7...v0.22.8)) so I think we can merge this now.

icweaver:
Thanks Chris for your guidance!

PR #60, opened 2025-02-21
Explicit imports

### LINKED ISSUE #55: Explicit imports
We are doing most of our importing from dependencies with generic `using` statements (e.g., `using Unitful`) which makes it hard for people not familiar with the dependencies (e.g., me) to tell where some methods/definitions are coming from. By using explicit imports or qualified names (e.g., `using Unitful: m, <etc...>` or `import Unitful as u; u.m` we make this discovery process easier  -- e.g., if I see an unqualified `interpolate` method being used in our code, I can go to the imports block and see where it is being imported from quickly. 

Similarly I would move import statements out of `include`d files like `color_law.jl` and `dust_maps.jl` and put all the imports in the same place at the module level in the `DustExtinction.jl` file. Having all the imports in one place simplifies the process of auditing the dependencies so you don't have to crawl through all the source files to find all the import statements.

I think this would improve maintainability / onboarding for new maintainers.

icweaver:
Closes #55

Took a pass at this. Can also apply this treatment to the tests if that is preferred. Update: actually, might be a good opportunity to migrate to https://github.com/julia-vscode/TestItems.jl


src/color_laws.jl
```diff
@@ -1,6 +1,3 @@
-using BSplineKit: interpolate, BSplineOrder, Natural
-using DelimitedFiles
```
icweaver, line 2: I think https://github.com/JuliaAstro/DustExtinction.jl/pull/54/files#diff-9880a77dc8047f68a334d57b03fdc90fa3d1fcc9898d313378cc414f32f077eeL265 makes this stale now, so I removed it. Not sure yet why Aqua is complaining about this

icweaver:
Hi Chris, sorry for letting this fall through the cracks. Things should be all green now and ready for your review

Issue #59, opened 2025-02-21
Towards Julia v1.9


icweaver:
Just opening this to track potential reasons for bumping our minimum supported Julia version from v1.6 to at least v1.9:

* Julia v1.10.9 is now the [current LTS](https://julialang.org/downloads/#long_term_support_release)
* [Package extensions](https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) require at least v1.9. This would be useful for #58
* [`Base.@kwdef`](https://docs.julialang.org/en/v1/base/base/#Base.@kwdef) is now exported in v1.9. This would remove our [Parameters.@with_kw](https://mauro3.github.io/Parameters.jl/stable/api/#Parameters.@with_kw-Tuple{Any}) dependency. Note: would need to work around `@assert` not being supported in `Base.@kwdef`, so probs not worth the hassle. [Related discussion](https://discourse.julialang.org/t/update-on-base-kwdef-versus-parameters-jl/95042/7)
    * Actually, maybe not that bad? #61 
* Apple silicon not supported on v1.6 or v1.7 https://discourse.julialang.org/t/how-to-fix-github-actions-ci-failures-with-julia-1-6-or-1-7-on-macos-latest-and-macos-14/117019

icweaver:
That would be my preference, but will defer to other folks that use this package more. I think @cgarling had thoughts about continuing to support 1.6 at some point

icweaver:
Was just reminded about this when going over #92. If we bump to 1.10, then we could drop stuff like this right?

https://github.com/JuliaAstro/DustExtinction.jl/blob/ffdb5b38fe88809d09d95051a878ca1d090355eb/test/dust_maps.jl#L48-L56

cgarling:
yeah the version check guards against earlier versions of Julia that don't support package extensions

icweaver:
Mkay, have a bit of downtime. Will start updating things

PR #61, opened 2025-02-21
Replace `Parameters.@with_kw` with `Base.@kwdef`

icweaver:
Partially addresses #59

Experimenting with replacing `Parameters.@with_kw` with `Base.@kwdef`


icweaver:
This should be ready for review too now. Thanks!

icweaver:
Sgtm!

Same. The only difference I am aware of is that [Parameters supports `@assert`s](https://discourse.julialang.org/t/update-on-base-kwdef-versus-parameters-jl/95042/7?u=icweaver), which one of our laws depended on: https://github.com/JuliaAstro/DustExtinction.jl/blob/2dad230d0d1097f1cbea51f0a42e94d072ea4876/src/fittable_laws.jl#L56-L57

There seems to be some discussion about [moving away from `@assert`](https://discourse.julialang.org/t/when-is-it-okay-to-use-assert/114399) these days though, so I don't feel too bad about not using it in this PR

cgarling:
Looks good to merge

PR #67, opened 2025-05-08
docs cleaned up

### LINKED ISSUE #43: Documentation for color laws
Currently the documentation for some of the color laws (e.g., F99, Cal00, CCM89) says that they return E(B-V). 
If I'm not mistaken, these actually give the ratio A_lambda / A_V, where A_lambda is the extinction in magnitudes at the input wavelength and A_V is the extinction in V band. A_V can be computed as R_v * E(B-V), where E(B-V) can be looked up via SFD98Map for Galactic extinction. Could you correct these docstrings?

icweaver:
Closes #43


src/color_laws.jl
```diff
@@ -37,7 +37,7 @@ bounds(::Type{OD94}) = (1000, 33333)
 
 Calzetti et al. (2000) Dust Law.
 
-Returns E(B-V) in magnitudes at the given wavelength. `λ` is the wavelength in Å
+Returns A(λ)/E(B-V) in magnitudes at the given wavelength. `λ` is the wavelength in Å
```
cgarling, line 40: I think this is A(lambda) / A_V as currently implemented. I like returning A(lambda) / A_V for consistency with other models.

Above equation 2 in Calzetti 2000 we have `k_full = A(lambda) / E(B-V)`.

The `k` we define on lines 57 and 59 is only the polynomial portion of the full `k` term. The full term would be `k_full = 2.659 * k + R_v` (see their equation 4). By the above, we have `k_full = A(lambda) / E(B-V) = 2.659 * k + R_v`. Therefore, what we return on line 61 is `1.0 + 2.659 * k / law.Rv == k_full / law.RV == A(lambda) / E(B-V) / law.RV == A(lambda) / A(V)`.

cgarling: I see you have it labeled as `A(x)/A(V)` in the plot above so this might be the only place it needs to be changed?

cgarling: I am using them interchangeably for simplicity but they are not in general the same. Part of the complexity is that the Calzetti 2000 law is technically an *attenuation* law rather than an *extinction* law (see [here](https://dust-attenuation.readthedocs.io/en/latest/dust_attenuation/attenuation.html)) although this distinction is not often discussed. Because C00 is measuring extragalactic attenuation, they employ different methods to constrain the observational `E(B-V)` values than studies looking at extinction in the MW. The nebular emission line tracers they use to measure `E(B-V)` values for the galaxies are summarized in Table 1 but presented in earlier work. The `E(B-V)` values measured in these nebular regions are systematically higher than the galaxy-wide average value, which is the quantity they actually want for predicting the attenutation of the stellar continuum (this is what they call `E_s(B-V)`) -- in this case, they say that the stellar continuum color excess `E_s(B-V)`) is 44% of the nebular color excess.

I think our implementation is fine because the conversion between the two has to be made by the user *after* we've provided them with `A(lambda) / A(V)`. In order to calculate a final attenuation in magnitudes `A(lambda)`, a user must determine what `A(V)` is for their region, which requires assuming an `E(B-V)` and an `RV`. It is at this point that the definition of the color excess becomes relevant, which is after we have already calculated `A(lambda) / A(V)`.

cgarling:
Looks good to merge

src/color_laws.jl
icweaver: Really interesting, thanks for taking the time to explain the distinction to me! That also clears up for me why Karl made both a `dust_extinction` and `dust_attenuation` package 😄

PR #85, opened 2026-03-30
Add dust map extension for SkyCoords

cgarling:
Adds SkyCoordsExt which allows querying SFD98Map with `SkyCoords.AbstractSkyCoords` via method`(m::SFD98Map)(s::SkyCoords.AbstractSkyCoord)`. I verified tests pass locally but I'm not able to build the docs locally (they stall out and do not finish) so idk if the docs are set up properly or not yet.


cgarling:
Docs are building, apparently just takes longer than I expected
https://juliaastro.org/DustExtinction.jl/previews/PR85/dust_maps/

abhro (approved):

PR #89, opened 2026-04-05
Allow precompiling in buildpkg step in CI

abhro:
Just to minimize log lines in the actual testing step, so that test failures are easier to inspect

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,8 @@ jobs:
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
+        with:
+          precompile: yes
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -38,6 +38,8 @@ jobs:
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
+        with:
+          precompile: yes
```
giordano, line 42: This step is useless and allowing precompilation is pointless and wasteful, as it'll precompile for a different configuration than the one used in CI, causing double work, not less

abhro: Ah :( I figured the main environment would contain most of the dependencies than the test, but invalidations do seem to be popping up everywhere

PR #92, opened 2026-04-08
Implement Chiang "Corrected SFD" map

cgarling:
This implements the "corrected SFD" map of [Chiang 2023](https://ui.adsabs.harvard.edu/abs/2023arXiv230603926C/abstract). Currently supports only the E(B-V) map and the mask, not the other data products (see https://zenodo.org/records/8207175). Basically everything you can do with SFD98Map you can also do with the new `CSFDMap` (integrations with Measurements.jl, SkyCoords.jl, Makie.jl are all included). This only does nearest pixel lookup (not interpolation) but this mirrors the behavior of python's `dustmap` package which doesn't do interpolation for this map either.

We might consider making an `AbstractEBVMap` type or something that SFD98 and CSFD could share so that we don't have to throw `m::Union{SFD98Map, CSFDMap}` on so many function arguments (e.g., `Makie.plottype`) but this would require an API design for what `AbstractEBVMap` should support which is larger than I want the scope of this PR to be. 

Edit:
Tests passed: https://github.com/JuliaAstro/DustExtinction.jl/actions/runs/24148410468
Docs build: https://juliaastro.org/DustExtinction.jl/previews/PR92/dust_maps/


icweaver (approved):
This looks great to me, thanks Chris!

I like the integration with Healpix.jl, and agree on punting design on an `AbstractEBVMap` to a future PR

cgarling:
Appreciate the review Ian!

PR #95, opened 2026-05-02
build: update from Julia v1.9 to v1.12

### LINKED ISSUE #59: Towards Julia v1.9

icweaver:
Closes #59

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -8,21 +8,23 @@ CairoMakie.activate!(type="png", px_per_unit=3)
 DocMeta.setdocmeta!(DustExtinction, :DocTestSetup, :(using DustExtinction); recursive = true)
 MakieExt = Base.get_extension(DustExtinction, :MakieExt)
 SkyCoordsExt = Base.get_extension(DustExtinction, :SkyCoordsExt)
-include("pages.jl")
 
 makedocs(;
     modules = [DustExtinction, MakieExt, SkyCoordsExt],
     sitename = "DustExtinction.jl",
     format = Documenter.HTML(;
-        prettyurls = get(ENV, "CI", nothing) == "true",
-        edit_link = "master",
         assets = [
             "assets/favicon.ico",
         ],
         canonical = "https://JuliaAstro.org/DustExtinction/stable/",
     ),
     authors = "Kyle Barbary, Mosé Giordano, Miles Lucas",
-    pages,
+    pages = [
+        "Home" => "index.md",
+        "Color Laws" => "color_laws.md",
+        "Dust Maps" => "dust_maps.md",
+        "Plotting" => "plotting.md",
+    ],
     warnonly = [:missing_docs],
 )
 
@@ -30,6 +32,5 @@ deploydocs(;
     repo = "github.com/JuliaAstro/DustExtinction.jl.git",
     push_preview = true,
     branch = "gh-pages",
-    devbranch = "master",
     versions = ["stable" => "v^", "v#.#"], # Restrict to minor releases
 )
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -13,6 +13,7 @@ makedocs(;
     modules = [DustExtinction, MakieExt, SkyCoordsExt],
     sitename = "DustExtinction.jl",
     format = Documenter.HTML(;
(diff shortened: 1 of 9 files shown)


docs/make.jl
```diff
@@ -8,28 +8,29 @@ CairoMakie.activate!(type="png", px_per_unit=3)
 DocMeta.setdocmeta!(DustExtinction, :DocTestSetup, :(using DustExtinction); recursive = true)
 MakieExt = Base.get_extension(DustExtinction, :MakieExt)
 SkyCoordsExt = Base.get_extension(DustExtinction, :SkyCoordsExt)
-include("pages.jl")
 
 makedocs(;
     modules = [DustExtinction, MakieExt, SkyCoordsExt],
     sitename = "DustExtinction.jl",
     format = Documenter.HTML(;
-        prettyurls = get(ENV, "CI", nothing) == "true",
```
cgarling, line 17: I would keep this, helps when building docs locally

icweaver: Yea, that used to be my thinking too. Here's [Michael's and Morten's take](https://github.com/JuliaDocs/Documenter.jl/pull/2402), which states the opposite. It seems preferable to just incorporate LiveServer.jl in the local doc dev process, then things are nice and uniform. I'd rather just point folks to this in the dev docs instead of promoting the old way, which seems to be falling into an anti-pattern these days. Does that sound reasonable to you?

cgarling: Last time I tried it I could not get LiveServer.jl to work for my configuration -- I do most of my dev work through WSL2 on Windows, so my Julia installation is on the Linux subsystem but I access .html files so on from a browser installed on the Windows side. I expect I had some issue with port forwarding or something between the Windows and Linux sides that prevented the Windows side from seeing the server hosted on the Linux side, yada yada the exact problem doesnt really matter I guess

My main point is this will make local dev harder for me (I will have to remember to change this back so I can look through the built docs properly) and unless it makes it easier for someone else who will be doing feature/docs work on this package (maybe you?) then I would leave it.

cgarling: An intermediate solution I just thought of is that we could use a new ENV variable to check whether to use prettyurl or not; that way I could define it locally and it would work but it would never get used otherwise, something like `prettyurls = get(ENV, "DOCUMENTER_PRETTYURL", nothing) == "true"` that would fix my particular problem.

icweaver: Ah, thank you for bringing the Windows issue to my attention, I was not aware. Happy to keep in the original CI check to accommodate.

I think your issue matters and would be good to investigate upstream, given how widely used LiveServer.jl is. Anyway, future issue for a future time, will revert here to maintain original behavior

icweaver: https://github.com/JuliaAstro/DustExtinction.jl/pull/95/commits/92a313a51713c2cfdb2a1fee4d64af5df65b9f57

icweaver:
Mkay, I think the basics bits for this PR are done here now and ready for a review

cgarling:
LGTM

